PR #329, opened 2018-09-25
fix storybook imports, use new components

Romanior:


Romanior:
as discussed naming is wrong

PR #337, opened 2018-10-09
Feature/update wallet face

Dragonesis:
- Update UI Wallets components
- Moving Wallets components to the root of components
- Updating components call path
- WalletCard add in storybook
- Add new prop for JFlatButton


src/components/WalletFace/WalletFace.js
```diff
@@ -38,7 +38,7 @@ const WalletFace = ({
               iconName='dots'
               iconSize='medium'
               iconColor='white'
-              hasNotBorder
+              isBordered={false}
```
IViol: boolean props should be false by default to avoid `isBordered={false}`
like any other html boolean attrs.

Dragonesis: Then I need to specify hasNotBorder or isNotBordered. Else the logic is broken.

src/routes/Wallets/routes/Create/components/WalletsCreateView.js
```diff
@@ -0,0 +1,63 @@
+// @flow
```
IViol: this file is not used

src/components/base/JIcon/JIcon.js
```diff
@@ -3,10 +3,8 @@
 import React, { PureComponent } from 'react'
 
 type Props = {
-  // @TODO: need to enumerate all icons
   name: string,
```
flexxnn: Please, create JIRA task for this issue (if it is not exists)

PR #338, opened 2018-10-09
Feature/small updates

flexxnn:


.flowconfig
```diff
@@ -1,7 +1,8 @@
 [ignore]
 .*/build/.*
 .*/src/static/.*
-.*/src/stories/.*
+.*/stories/.*
+*.old.js
```
IViol: why we need *.old.js?

PR #345, opened 2018-10-11
Update delete wallet page

IViol:
JWWEB-95 Delete wallet page


flow-typed/types.js
```diff
@@ -433,13 +428,12 @@ declare type RouterData = {
  */
 declare type State = {|
   +wallets: WalletsState,
-  +networks: NetworksData,
-  +mnemonicAddresses: MnemonicAddressesData,
   +walletsCreate: WalletsCreateState,
   +walletsImport: WalletsImportState,
+  +networks: NetworksData,
+  +mnemonicAddresses: MnemonicAddressesData,
```
flexxnn: We are going to rename XXXData -> XXXState

src/routes/Wallets/routes/Delete/sagas/walletsDelete.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+
+import { delay } from 'redux-saga'
+import { put, takeEvery } from 'redux-saga/effects'
+
+import config from 'config'
+import walletsWorker from 'workers/wallets'
+import * as wallets from 'routes/Wallets/modules/wallets'
```
Romanior: would be nice to have absolute or relative paths, but not a mix of both

Romanior: let's use relative inside module, absolute for the rest

PR #346, opened 2018-10-11
fix absent transactions reducers

flexxnn:


src/store/reducers.js
```diff
@@ -26,6 +27,7 @@ export function makeRootReducer(asyncReducers: ?Reducers): Reducer<any, any> {
     wallets,
     walletsCreate,
     digitalAssets,
+    transactions,
```
IViol: if so, please remove it's injecting inside transactions routes

flexxnn: With clean installation application crashes without it

flexxnn: Actually we have broken solution in develop

PR #349, opened 2018-10-12
fix flowtypes

flexxnn:
render and placeholder can be null (defaultProps), fix flow types

### DIFF
--- a/src/components/base/JInput/JInput.js
+++ b/src/components/base/JInput/JInput.js
@@ -9,7 +9,7 @@ type Props = {
   onChange: ?((string) => void),
   name: ?string,
   label?: string,
-  placeholder?: string,
+  placeholder: ?string,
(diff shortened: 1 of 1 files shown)


src/components/base/JInput/JInput.js
```diff
@@ -20,7 +20,7 @@ type Props = {
   isDisabled: boolean,
   isPinCode: boolean,
   isMultiline: boolean,
-  render?: Function
+  render: ?Function
```
IViol: Do we need these props in defaultProps? Or they should be removed?

flexxnn: When we have default prop with possible null value we should use "?Function" here
And, according this documentation https://flow.org/en/docs/react/components/#toc-using-default-props we don't need to use "render? " wen we have default prop

PR #352, opened 2018-10-12
Update package lock

IViol:


Romanior:
de will remove it for now from github untill development process is set and we will start deployments.

PR #372, opened 2018-10-25
Merge with feature/assets

Dragonesis:


src/components/DigitalAssetsGrid/digitalAssetsGrid.scss
```diff
@@ -3,34 +3,21 @@
...
-      height: 1px;
-      background: rgba($white, 0.15);
-      content: '';
-    }
-  }
-
-  @media only screen and (max-width: 1400px) {
-    grid-template-columns: 1fr 1fr;
+  > .box {
+    width: 33.333%;
```
IViol: we have fixed width for the digital asset item 320px

Dragonesis: This is not true. Cards are resized to fit the container. This can be seen in the task task-11-adaptation file

PR #393, opened 2018-11-06
Flowtype fixes

flexxnn:


flow-typed/types-digitalAssets.js
```diff
@@ -0,0 +1,54 @@
+// @flow
```
IViol: I suggest to place these files inside `flow-typed/redux` dir
so I would like to rename `flow-typed/types-digitalAssets.js` -> `flow-typed/redux/digitalAssets.js`

`types-digitalAssets` name looks really strange with dash and camelCase

flow-typed/types-digitalAssets.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Digital assets
```
IViol: When we have name of specific state in file, such comments become useless

PR #396, opened 2018-11-09
Feature/update JTooltip

Dragonesis:


src/components/base/JTooltip/JTooltip.js
```diff
@@ -1,30 +1,40 @@
...
+            <JText
+              value={text}
+              size='small'
+              color='white'
+              weight='bold'
+            />
+          )}
+          placement='top'
+        >
+          <div>{children}</div>
```
IViol: > <div>{children}</div>

it could be just {children}

IViol: we need to avoid to use redundant DOM nodes, if it is not used for markup/styling

Dragonesis: In this case doesn't work tooltip

PR #397, opened 2018-11-09
Feature/add JSwitch

Dragonesis:


src/components/base/JSwitch/jSwitch.scss
```diff
@@ -0,0 +1,57 @@
...
+        height: 20px;
+        position: absolute;
+        border-radius: 100px;
+        background-color: #fff;
+        will-change: transform;
+        transition: transform 0.3s ease;
+      }
+    }
+
+    > .checkbox:checked ~ .switch {
```
IViol: It's ok, but not so obviously.
For example, some developer will need to make wrapper for `.switch` (for animation or any other purposes), and behaviour of the component will be broken, if he forget to change css.
We have controlled by js prop `isChecked` so let's use it to toggle some parent class (modifier, e.g. `-checked`)
```
<div className={classNames('j-switch', isChecked && '-checked')}>
```

src/components/base/JSwitch/jSwitch.scss
```diff
@@ -0,0 +1,57 @@
...
+
+      &::after {
+        content: '';
+        top: 2px;
+        left: 2px;
+        width: 20px;
+        height: 20px;
+        position: absolute;
+        border-radius: 100px;
+        background-color: #fff;
```
IViol: background-color: $white;

src/components/base/JSwitch/JSwitch.js
```diff
@@ -0,0 +1,45 @@
...
+
+import React, { PureComponent } from 'react'
+
+type Props = {|
+  +onChange: ?((boolean) => void),
+  +name: string,
+  +isChecked: boolean,
+|}
+
+class JCheckbox extends PureComponent<Props> {
```
flexxnn: Rename to JSwitch

PR #402, opened 2018-11-14
Assets manager

flexxnn:


flow-typed/components.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+
+declare type DigitalAssetsGridItemType = {
```
IViol: DigitalAssetsGridItem~~Type~~

src/routes/DigitalAssets/routes/Manage/modules/digitalAssetsManage.js
```diff
@@ -0,0 +1,71 @@
...
+
+      return {
+        ...state,
+        searchQuery: query,
+      }
+    }
+
+    case CLEAN: {
+      return {
+        ...initialState,
```
IViol: just `return initialState`

PR #412, opened 2018-11-22
JWWEB-230: Fail build hard

stcherenkov:
Sometimes build can fail deep inside webpack, which does not finish all loops properly and prevents process from exit. We are now calling `process.abort` that closes process "unsafe", but **for sure**.

Also removed build scripts from lint staged list.

You should probably merge this branch into your WIP-branches to stop PRs from hanging up Jenkins.


scripts/build.js
```diff
@@ -98,7 +98,7 @@ measureFileSizesBeforeBuild(paths.appBuild)
     err => {
       console.log(chalk.red('Failed to compile.\n'));
       printBuildError(err);
-      process.exit(1);
+      return process.abort();
```
IViol: `return` useless here because process will shutdown immediately

stcherenkov: Thought the same first, but without `return` it doesn't.

¯\_(ツ)_/¯

PR #413, opened 2018-11-22
JWWEB-232 SettingsCard implementation

Creepypoke:
SettingsCard component implementation.

### DIFF
--- a/src/components/DigitalAssetsGrid/DigitalAssetsGrid.js
+++ b/src/components/DigitalAssetsGrid/DigitalAssetsGrid.js
@@ -25,7 +25,6 @@ class DigitalAssetsGrid extends PureComponent<Props> {
 
     return (
       <div className='digital-assets-grid'>
-        {items.length === 0 && <Empty />}
         {items.map(({ asset, balance }) => (
           <div className='box' key={asset.address}>
             { /* @TODO: add fiatCurrency, fiatBalance */ }
@@ -40,11 +39,18 @@ class DigitalAssetsGrid extends PureComponent<Props> {
             />
           </div>
         ))}
-        {items.length &&
+        {items.length ? (
           <div className='box'>
             <AddDigitalAsset onClick={addAssetClick} />
           </div>
-        }
+        ) : (
+          <Empty
+            image='screen-search'
+            description='There are no Digital Assets to show'
+            color='gray'
+            isTransparent
+          />
+        )}
       </div>
     )
   }
--- a/src/components/DigitalAssetsGrid/Empty/Empty.js
+++ b/src/components/DigitalAssetsGrid/Empty/Empty.js
@@ -1,25 +1,40 @@
 // @flow
 
-import React from 'react'
+import React, { PureComponent } from 'react'
 
 import JThumbnail from 'components/base/JThumbnail'
 
-const DigitalAssetsEmpty = ({ color }: Props) => (
-  <div className='digital-assets-empty'>
-    <JThumbnail
-      image='cloud'
-      color={(color === 'white') ? 'gray' : 'white'}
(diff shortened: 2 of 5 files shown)


src/components/SettingsCard/SettingsCard.js
```diff
@@ -0,0 +1,56 @@
...
+  iconName?: string,
+  iconColor?: 'white' | 'blue' | 'gray' | 'sky',
+  iconSize?: 'small' | 'medium' | 'large' | 'xlarge',
+}
+
+SettingsCard.defaultProps = {
+  iconName: 'setting',
+  iconColor: 'gray',
+  iconSize: 'medium',
+}
```
IViol: if icon props could be excluded from component props, why we set some default values for them?
all of them could be null.
And before using of them, we should check `iconName && iconColor && iconSize && <...>`

src/components/SettingsCard/SettingsCard.js
```diff
@@ -0,0 +1,56 @@
...
+        />
+      </div>}
+      <div className='summary'>
+        <div className='title'>
+          <JText value={title} color='dark' weight='bolder' size='header' />
+        </div>
+        {description &&
+        <div className='description'>
+          <span className='text'>{description}</span>
+        </div>}
```
IViol: description should be wrapped in JText

src/components/SettingsCard/settingsCard.scss
```diff
@@ -0,0 +1,50 @@
+.settings-card {
+  all: initial;
+
+  @include box;
+  @include font;
+  @include flex-column;
+  @include hover-animation;
+
```
IViol: you don't need to include font mixin if description will be wrapped by JText

stories/components/SettingsCrad.stories.js
```diff
@@ -0,0 +1,91 @@
...
+          max-width: 1020px;
+          margin: 0 auto;
+          display: flex;
+          flex-wrap: wrap;
+        }
+        .settings-grid > .box {
+          width: 33.333%;
+          padding: 0 10px 20px 10px;
+        }`}
+      </style>
```
IViol: better to write css in `stories/styles.css`

src/components/SettingsCard/SettingsCard.js
IViol: but if we don't have setting cards without icons, we should make these props required

src/components/SettingsCard/SettingsCard.js
```diff
@@ -0,0 +1,56 @@
...
+import { JIcon, JText } from 'components/base'
+
+type Props = {
+  title: string,
+  description: string,
+  path: string,
+  iconName?: string,
+  iconColor?: 'white' | 'blue' | 'gray' | 'sky',
+  iconSize?: 'small' | 'medium' | 'large' | 'xlarge',
+}
```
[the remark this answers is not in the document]
flexxnn: For better type checking it is good to use strict operator "|", and make props read only
`type Props = {|
  +title: string,
  +description: string,
  +path: string,
  +iconName: string,
  +iconColor: 'white' | 'blue' | 'gray' | 'sky',
  +iconSize: 'small' | 'medium' | 'large' | 'xlarge',
|}`

src/components/SettingsCard/SettingsCard.js
flexxnn: I think, here the `size` is always medium. Isn't it?

Creepypoke: Both of you are right: `iconName` is should be required, and `iconSize` always is should has `medium` value, only color of icon has options.

PR #416, opened 2018-11-23
Update .gitignore

stcherenkov:
Motivation: there are shared parts of IDEs / editors config that should not be ignored

Include OS-specific template for MacOS and Windows
Include VSCode and WebStorm templates that exclude personal configurations, but leave shared params in repo
Update NodeJS ignore config with official template

For list of templates see [github/gitignore](https://github.com/github/gitignore)


Creepypoke:
I agree this idea, but I want to hear the opinion of regular team

PR #421, opened 2018-11-26
Implement transactions management

IViol:


src/routes/sagas/transactions.js
```diff
@@ -252,7 +268,7 @@ function* recursiveRequestTransactions(
       })
 
       if (fromBlockNew > 0) {
-        yield call(delay, 100)
+        yield call(delay, 1000)
```
flexxnn: Why we need this delay here?

PR #424, opened 2018-11-27
update transactions by asset view

IViol:


src/services/blockExplorer.js
```diff
@@ -125,11 +125,10 @@ function filterETHTransactions(list: Array<any>): Array<Object> {
     }
 
     const {
-      amount,
       contractAddress,
     }: Object = item
```
flexxnn: Change Object here

src/services/blockExplorer.js
```diff
@@ -125,11 +125,10 @@ function filterETHTransactions(list: Array<any>): Array<Object> {
...
 
-    const isEmptyAmount: boolean = (amount === 0)
+    const isEmptyAmount: boolean = (Number(item.value) === 0)
```
flexxnn: `new BigNumber((item.value).eq(0)`

PR #429, opened 2018-11-29
JWWEB-218 Internet connection error

Dragonesis:


src/components/base/JThumbnail/jThumbnail.scss
```diff
@@ -16,13 +16,13 @@
     @include flex-column-center;
 
     padding-top: 10px;
-    line-height: 24px;
+    line-height: (18/11);
```
IViol: we must specify line-height in pixels

Dragonesis: Line height in pixels is bad practice

PR #430, opened 2018-11-30
blocks management refactoring

IViol:


src/components/DigitalAssetsManage/DigitalAssetsManage.js
```diff
@@ -1,42 +1,52 @@
 // @flow
 
 import React, { PureComponent } from 'react'
-import { handle } from 'utils/eventHandlers'
+
+import handle from 'utils/eventHandlers/handle'
```
stcherenkov: These "handle" thing looks old enough, so I don't think it blocks PR from merging. But it would be much better if it was called `ignoreEventHandle` or something else more detailed than now.

src/config.js
```diff
@@ -138,4 +140,19 @@ export default {
     '42': 'api-kovan',
     '4': 'api-rinkeby',
   },
+
+  /**
+   * latest block sync timeout
```
stcherenkov: Not sure if we should write comments like this. Variable name is already explaining its purpose.

src/routes/DigitalAssets/routes/Grid/DigitalAssetsGridView.js
```diff
@@ -18,15 +22,14 @@ const DIGITAL_ASSETS_TABS = {
 type Props = {|
   openView: () => void,
   closeView: () => void,
-  setSearchQuery: (string) => void,
+  addAssetClick: () => void,
```
stcherenkov: Future suggestion: there is could rule in eslint react plugin that suggests using `handle` (or any other) prefix to differ between DOM event handlers and other methods. Maybe we should consider using it?

src/components/DigitalAssetsManage/DigitalAssetsManage.js
IViol: we already have `ignoreEvent` util method that calls preventDefault and then stopPropagation for event (to prevent bubbling). But we can rename it to `preventEvent`, and this method `handle` to `ignoreEvent`.

Actually I don't sure that its name should be so detailed, because it says `handle` ( `that function` ) ( `with these args` )

PR #431, opened 2018-11-30
WEB-2: update build process

stcherenkov:
We update for most recent versions of webpack, babel etc., based on create-react-app latest webpack config saving our additions.

What should be noted:
- environment configs moved to external .env files and are loaded via `env-cmd`
- new `/reports` dir will collect all the code metrics (for now it is only _flow coverage report_ and _webpack bundle analyzer report_)
- we moved to having one webpack config for both production and development environments, meaning simpler support and update
- new `/docs` dir will collect all the autogenerated docs (for now it is storybook static build, but we could also store any other docs, i.e. flow-typings-based)

Also there is new webpack config, but it is mostly self-explanatory.


webpack.config.js
```diff
@@ -0,0 +1,412 @@
...
+
+    // Makes some environment variables available to the JS code, for example:
+    // if (process.env.NODE_ENV === 'production') { ... }.
+    // It is absolutely essential that NODE_ENV was set to production here.
+    // Otherwise React will be compiled in the very slow development mode.
+    new webpack.DefinePlugin({
+      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
+      '__DEV__': process.env.NODE_ENV === 'development',
+      '__PROD__': process.env.NODE_ENV === 'production',
+      'process.env.BUILD_NUMBER': process.env.BUILD_NUMBER || 'dev',
```
IViol: we can use existing flags here isEnvProduction & isEnvDevelopment

webpack.config.js
```diff
@@ -0,0 +1,412 @@
...
+// `publicUrl` is just like `publicPath`, but we will provide it to our app
+// as %PUBLIC_URL% in `index.html` and `process.env.PUBLIC_URL` in JavaScript.
+// Omit trailing slash as %PUBLIC_URL%/xyz looks better than %PUBLIC_URL%xyz.
+const publicUrl = publicPath.slice(0, -1)
+
+// This is the production configuration.
+// It compiles slowly and is focused on producing a fast and minimal bundle.
+// The development configuration is different and lives in a separate file.
+module.exports = {
+  mode: isEnvProduction ? 'production' : isEnvDevelopment && 'development',
```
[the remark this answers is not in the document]
stcherenkov: In this case webpack would use default value – `none`, so it's mostly ok

webpack.config.js
stcherenkov: thnx, fixed

stcherenkov:
Heads up! I removed `window` call inside web worker in HMR mode and checked Docker container deployment.

stcherenkov:
Added gzip compression to build config. See `CompressionPlugin` in `/webpack.config.js`.

We should also add brotli support, but not today. ¯\\\_(ツ)\_/¯

webpack.config.js
```diff
@@ -0,0 +1,432 @@
...
+  module: {
+    strictExportPresence: true,
+
+    rules: [
+      // Disable require.ensure as it's not a standard language feature.
+      { parser: { requireEnsure: false } },
+
+      // First, run the linter.
+      // It's important to do this before Babel processes the JS.
+      {
```
flexxnn: I sggest to disable eslint formatter, because we have husky + eslint and this is not nesessuary

stcherenkov: Fixed. We show errors only in production now.

PR #437, opened 2018-12-04
WEB-4: Set up simple smoke testing via BrowserStack

stcherenkov:
We now have webdriver.io config that loads our jasmine tests through specified BrowserStack browsers.

It works, except for iPad, that somehow breaks connection with BrowserStack server forever.

There is also no sane way to capture logs on test fail. We should wait until webdriverio@5 is out, maybe it will be fixed (according to sources it is very likely).

Right now test fails in IE and Opera both in development and production. Most likely it is because of some ES6-syntax that is not transpiled to older JS. Needs further investigation.

To start testing you need to specify two environment variables: `BROWSERSTACK_USERNAME` and `BROWSERSTACK_ACCESS_KEY`. Feel free to contact me, if you don't know where to get those.

If you are running Docker container or on non-standard port, you could specify your server port with `TEST_SERVER_PORT` env var.

Shorthand for running all tests is `npm test` (there are only smoke tests to run right now).

Reports are collected at `/reports/browserstack/*.xml`, where wildcard is browser name, run counter etc.


stcherenkov:
Oh, you also need to install [BrowserStackLocal app](https://www.browserstack.com/local-testing) to run tests on local server.

stcherenkov:
And here is the [Jenkins Plugin](https://www.browserstack.com/automate/jenkins).

stcherenkov:
To save vital testing information for future generations, I added `/docs/testing.md`.

wdio.conf.js
```diff
@@ -0,0 +1,120 @@
...
+      os_version: '10',
+      browserName: 'Edge',
+    },
+    {
+      os: 'Windows',
+      os_version: '10',
+      browserName: 'Firefox',
+    },
+    {
+      os: 'Windows',
```
edanilov: is needed to add Opera browser for Win 10

wdio.conf.js
```diff
@@ -0,0 +1,120 @@
...
+  suites: {
+    smoke: [
+      './tests/initial-load.spec.js',
+    ],
+  },
+
+  maxInstances: 1,
+  capabilities: [
+    // {
+    //   device: 'iPad 6th',
```
edanilov: is needed to add some Android tablets (UC Browser etc)

wdio.conf.js
stcherenkov: BrowserStack does not support automate for Opera on Windows, sorry.

wdio.conf.js
stcherenkov: Fixed

PR #438, opened 2018-12-05
WEB-3: unit testing environment

stcherenkov:
This PR finishes establishment of testing infrastructure.

It features updated jest config and a simple test just to check if it works all together.

If you run into cryptic `[BABEL]` errors when you `npm run test:unit`, try:

```
npm r jest && npm i -D jest
```

It links then via babel-core@bridge instead of internal babel-core@6.

See also https://github.com/jibrelnetwork/jwallet-web/pull/437 for functional testing configuration.


src/components/base/JText/tests/JText.test.js
```diff
@@ -0,0 +1,15 @@
...
+
+describe('JText', () => {
+  test('is available', () => {
+    expect(JText).toBeDefined()
+  })
+
+  test('renders', () => {
+    const wrapper = shallow(<JText value='text' />)
+    expect(wrapper.text()).toBe('text')
+  })
```
Creepypoke: Beautiful!

PR #440, opened 2018-12-06
transactions by range management

IViol:
- flow types for config.js
- fix BigNumber types
- fix transactions types
- implement storing of transactions by block number (`toBlock` number)
- fix scheduler types
- update displaying of ETH transactions with deploying of contract
- add sorting (by block timestamp) of transactions in UI
- add initialising of transactions with empty structures at start of syncing
- update fetching of transactions
- implement re-syncing of transactions if some requests was failed or not finished


src/utils/numbers/isZero.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+import { BigNumber } from 'bignumber.js'
+
+function isZero(amount: string): boolean {
```
flexxnn: Here, we can use `string | number | BigNumber`

src/utils/type/index.js
```diff
@@ -1,11 +1,13 @@
 // @flow
 
+import isVoid from './isVoid'
 import isArray from './isArray'
 import isObject from './isObject'
 import isString from './isString'
 import isNumber from './isNumber'
```
stcherenkov: Lodash has all of these, btw

src/utils/numbers/isZero.js
IViol: fixed [here](https://github.com/jibrelnetwork/jwallet-web/pull/441/commits/8ba1ac5589d08555665366aaa738b04ec34e6a2b)

src/utils/type/isVoid.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+
+function isVoid(data: any): boolean {
+  return (data == null)
```
Creepypoke: Why are you using round brackets?
Using the abstract equality (`==`) operator is deliberate?

IViol: (data == null) checks null or undefined

src/utils/transactions/getAddressLink.js
```diff
@@ -2,8 +2,10 @@
 
 import config from 'config'
 
-function getAddressLink(address: Address, blockExplorerSubdomain: string): string {
-  return `//${blockExplorerSubdomain}.${config.blockExplorerLink}/address/${address}`
+function getAddressLink(address: ?Address, blockExplorerSubdomain: string): string {
+  return address
+    ? `//${blockExplorerSubdomain}.${config.blockExplorerLink}/address/${address}`
+    : ''
```
Creepypoke: I think, we can get strict `Address` type as parameter without nullable state, and check nullable state of parameter before call this function.

src/utils/digitalAssets/parseBalance.js
```diff
@@ -1,16 +1,9 @@
...
 import { divDecimals } from '../numbers'
 
-function parseBalance(
-  value: string,
-  decimals: number = 18
-): string {
-  const normalized = divDecimals(new BigNumber(value), decimals)
-
-  return normalized.toString()
+function parseBalance(value: string, decimals: number = 18): string {
```
Creepypoke: What does number 18 mean? Maybe, we can set constants for "Magical numbers"?

PR #442, opened 2018-12-10
WEB-7 Implementation of search on the settings page

Creepypoke:
Search field was added to Settings page


src/utils/formatters/formatCurrency.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import { CURRENCIES_MAP } from 'data/settings'
+import { CURRENCIES_MAP } from '../../data/settings'
```
stcherenkov: ?

src/utils/formatters/formatLanguage.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import { LANGUAGES_MAP } from 'data/settings'
+import { LANGUAGES_MAP } from '../../data/settings'
```
stcherenkov: ?

src/routes/Settings/SettingsIndexView.js
```diff
@@ -35,123 +162,48 @@ class SettingsIndexView extends PureComponent<Props, *> {
...
+      localCurrencyCode,
+      defaultGasPrice,
+      systemLanguageCode,
+      hasPinCode,
+      walletName,
+      networkName,
+      isFullMnemonic,
+      derivationPath,
+      passphrase,
+    }).filter((elementProps) => {
```
stcherenkov: Most of the times users would try to find settings by non-trivial queries. For example, users can look for "language" using "locale" or "translation".

For those cases consider to add smth like `searchTags` field, empty and ignored by default. We will fill it with keywords as soon as we get our hands on real-world data.

Creepypoke: Meh...

src/routes/Settings/SettingsIndexView.js
```diff
@@ -2,20 +2,147 @@
...
+      searchQuery: '',
+    }
+  }
+
+  setSearchQuery = (query: string): void => {
+    this.setState({ searchQuery: query.toLowerCase() })
+  }
+
+  filterCardByQuery = (query: string): boolean => {
+    const queryPattern = new RegExp(`.*${this.state.searchQuery}.*`)
```
stcherenkov: This regexp is very inefficient. You don't need leading and trailing `.*`, but you need flags `g` and `i`.

stcherenkov: If you add `i`, you also don't need `toLowerCase` conversion.

PR #467, opened 2018-12-21
Feature/update double input

Dragonesis:


src/components/DoubleInput/DoubleInput.js
```diff
@@ -19,80 +18,68 @@ type Props = {|
...
+  +onChangeRight: ?((value: string) => void),
   +onClose: ?(() => void),
-  +valueFiat: string,
-  +valueAmount: string,
-  +errorMessageAmount: string,
+  +valueLeft: string,
+  +valueRight: string,
+  +placeholderLeft: string,
+  +placeholderRight: string,
+  +errorMessage: string,
```
stcherenkov: `errorMessage` is always for both inputs?

src/components/DoubleInput/DoubleInput.js
```diff
@@ -19,80 +18,68 @@ type Props = {|
...
   +isLoading: boolean,
```
stcherenkov: Loading can happen for both inputs only?

src/components/DoubleInput/DoubleInput.js
```diff
@@ -1,8 +1,7 @@
 // @flow
 
-import React, { PureComponent, Fragment } from 'react'
-import classNames from 'classnames'
-import { JInput, JText, JLoader, JIcon } from 'components/base'
+import React, { PureComponent } from 'react'
+import { JInput, JLoader, JIcon } from 'components/base'
 
 /*
```
flexxnn: remove it

src/components/DoubleInput/DoubleInput.js
```diff
@@ -19,80 +18,68 @@ type Props = {|
...
+  +errorMessage: string,
   +isLoading: boolean,
-  +isActive: boolean,
 |}
 
 class DoubleInput extends PureComponent<Props> {
   static defaultProps = {
     isLoading: false,
-    isActive: false,
     onClose: null,
```
flexxnn: add here
`onChangeRight: null`

src/components/DoubleInput/DoubleInput.js
```diff
@@ -19,80 +18,68 @@ type Props = {|
...
-                    <JLoader color='blue' />
-                  </div>
-                )}
-              </Fragment>
-            )}
-          </div>
+          )}
         </div>
-        {onClose &&
+        {!onClose &&
```
flexxnn: close should only appears when onClose is exists in the props

src/components/DoubleInput/DoubleInput.js
flexxnn: Yes, and I think we should add the infoMessage (yellow message) here. We need when gasPrice value is too small

src/components/DoubleInput/DoubleInput.js
flexxnn: In design we have only one loader, I think in send asset form thereis only one place with this loader (when fiat conversion rate is loading)

src/components/DoubleInput/DoubleInput.js
```diff
@@ -28,45 +32,63 @@ class DoubleInput extends PureComponent<Props> {
...
+        {onClose &&
         <div className='close' onClick={onClose}>
           <JIcon name='padding-cross' size='medium' color='gray' />
         </div>}
+        {errorMessage &&
+        <div className='message'>
+          <JText value={errorMessage} color='red' size='small' />
+        </div>
+        }
+        {warningMessage &&
```
flexxnn: I think, error has biggest priority than warning, when we have both messages

`warningMessage && !errorMessage `

PR #468, opened 2018-12-21
Cover favorites edge cases

stcherenkov:
Add empty list management
Add link to add favorite

### DIFF
--- a/src/routes/Favorites/FavoritesIndexView.js
+++ b/src/routes/Favorites/FavoritesIndexView.js
@@ -1,18 +1,21 @@
 // @flow
 
 import React, { PureComponent } from 'react'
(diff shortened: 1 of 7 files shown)


src/routes/Favorites/FavoritesIndexView.js
```diff
@@ -68,21 +71,69 @@ class FavoritesIndexView extends PureComponent<Props, State> {
...
+                />
+              </div>
+            )
+          })}
+        </div>
+      )
+    }
+
+    return (
+      <div className='overlaycontainer'>
```
IViol: would be better `container -overlay`

stcherenkov: Fixed

PR #474, opened 2018-12-26
digital assets send refactoring

IViol:


IViol:
@stcherenkov these changes (refactoring) was part of implemented features.
Before, code was in unsupportable and almost non-working condition, so that, refactoring(and fixing) was necessary.

So big PR, because a lot of sub-PRs were based on it. Because we didn't have resources to review it in time.

PR #475, opened 2018-12-26
JWWEB-270 Responsive condition of layout was removed

Creepypoke:


src/styles/_base.scss
```diff
@@ -4,6 +4,11 @@
 @import 'tokens';
 @import 'zindex';
 
+/**
+ * Window
+ */
+$window-min-width: 900px;
```
IViol: as I remember, we should limit width size to 1024px

src/public/index.html
```diff
@@ -18,7 +18,7 @@
     <meta name='og:site_name' content='jWallet by the Jibrel Network' />
     <meta name='og:locale' content='en_US' />
     <meta name='og:type' content='website' />
-    <meta name='viewport' content='width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no' />
+    <meta name='viewport' content='initial-scale=1.0, maximum-scale=1.0, user-scalable=no' />
```
IViol: I think it should be just
`<meta name='viewport' content='width=1024' />`

PR #477, opened 2018-12-27
JWWEB-399 Make Ethereum asset always active

Dragonesis:


src/components/AssetItem/AssetItem.js
```diff
@@ -167,6 +168,7 @@ class AssetItem extends PureComponent<Props, StateProps> {
                 <JSwitch
                   onChange={setIsActive}
                   isChecked={isActive}
+                  isDisabled={isActive ? checkETH(address) : false}
```
Creepypoke: What about `checkETH(address) || isActive` ?

stcherenkov:
I suggest we also add link from balance to assets list (`/digital-assets/grid`)

stcherenkov:
This PR actually consists of two, right? JWWEB-399 and displaying ETH balance in MenuPanel?

src/components/AssetItem/AssetItem.js
Creepypoke: I think we need that Ethereum assets will be active all time, or not?

IViol: we should lock Ethereum to disable when it is already active

PR #478, opened 2018-12-28
Implement transactions comments management

IViol:


src/routes/modules/comments.js
```diff
@@ -0,0 +1,51 @@
+// @flow
+
+export const EDIT = '@@comments/EDIT'
+export const SET_ITEMS = '@@comments/SET_ITEMS'
+
+export function edit(id: CommentId, value: string) {
```
stcherenkov: And what about `delete`?

IViol: to remove comment we should pass value with empty string, I will describe why so a bit more later

PR #481, opened 2018-12-28
Settings list view fix for different types of wallet

Creepypoke:


src/routes/Settings/SettingsIndexView.js
```diff
@@ -176,7 +176,7 @@ class SettingsIndexView extends PureComponent<Props, State> {
     }).filter((elementProps) => {
       const { description, title, searchTags } = elementProps
 
-      if (description === null || description === false) {
+      if (description === null || description === '') {
```
stcherenkov: `if (!description)`?

stcherenkov: Why do you set it to empty string and then check for null?

stcherenkov: Are you sure it is a good idea to check setting display implicitly via its description, not via specific property?

IViol: also
`if (description === null || description === '')` equals `if (!description)`

src/routes/Settings/SettingsIndexView.js
```diff
@@ -71,13 +71,13 @@ const getSettingsCardProperties = ({
   searchTags: '',
 }, {
   title: 'Sign a message',
-  description: isFullMnemonic ? 'Enable' : ' ',
+  description: isFullMnemonic ? 'Enable' : '',
   path: 'settings/sign',
   iconName: 'message',
```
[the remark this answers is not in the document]
Creepypoke: Thanks! Please, where I can found some information about accessibility settings for different types of wallet ?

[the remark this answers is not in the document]
IViol: Sorry, there is no place with such info. You just need to know what actions user can do with mnemonic, private key and basic ETH address.
For example, diff between address and private key is that user can sign transaction (actually any sort of data) only when he has private key.
And about mnemonic attributes you can read from BIPs, that I shared.

src/routes/Settings/SettingsIndexView.js
```diff
@@ -82,7 +83,7 @@ const getSettingsCardProperties = ({
...
   searchTags: '',
-  isVisible: isFullMnemonic,
+  isVisible: isFullMnemonic || customType === 'privateKey',
```
IViol: > isFullMnemonic || customType === 'privateKey'

too verbose...
it is just `!isReadOnly`

Creepypoke: why we have two properties: `custom Type` and `type`? What's difference between them?

PR #486, opened 2018-12-28
JWWEB-278: add empty route for wallet verification page

stcherenkov:
View logic will be implemented in another PR

### DIFF
--- a/src/routes/Verify/verifyView.scss
+++ b/src/routes/Verify/verifyView.scss
@@ -1,0 +1,75 @@
+@import 'styles/mixins';
+
+.verify-view {
+  all: initial;
+
+  @include box;
+  @include flex-column-center;
+
+  position: relative;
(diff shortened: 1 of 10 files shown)


src/routes/Verify/verifyView.scss
```diff
@@ -0,0 +1,75 @@
...
+  all: initial;
+
+  @include box;
+  @include flex-column-center;
+
+  position: relative;
+  width: 100%;
+  height: 100%;
+
+  > .header {
```
flexxnn: Seems, it is not used now?

stcherenkov: Yep, copy-pasted from another view.

But as soon as we add smth of value here we *may* need to copy-paste it again.

I think it is better to clean up after actual implementation.

PR #488, opened 2018-12-29
fix etherscan endpoint

IViol:


src/services/blockExplorer.js
```diff
@@ -31,8 +31,8 @@ function getMainEthersanAPIEndpoint(apiSubdomain: BlockExplorerAPISubdomain): st
   const isMainNetwork: boolean = (apiSubdomain === MAIN_NETWORK_ETHERSCAN_API_SUBDOMAIN)
 
   return (__DEV__ && isMainNetwork)
-    ? 'https://ropsten.etherscan.io/api'
-    : `https://${apiSubdomain}.etherscan.io/api`
+    ? 'http://ropsten.etherscan.io/api'
+    : `http://${apiSubdomain}.etherscan.io/api`
```
stcherenkov: Please add `// FIXME` comment or task / bug so we don't forget to update those before release

PR #492, opened 2019-01-09
Remove unused sidebar buttons

stcherenkov:


src/components/MenuPanel/WalletManager/Details/Actions/Actions.js
```diff
@@ -12,21 +12,11 @@ function MenuPanelWalletManagerDetailsActions() {
...
-      <MenuPanelActionsItem
-        icon='lock'
-        path='#'
-        label='Lock the app'
-      />
       <MenuPanelActionsItem
         icon='plus'
-        path='#'
+        path='/wallets/create'
         label='Add new wallet'
```
IViol: Im not sure that 'Add new wallet' is only `create` action, but not `import` also.

stcherenkov: Yep, but it could not be both. :)

PR #493, opened 2019-01-09
Add pending transactions availability check

stcherenkov:


stcherenkov:
Fixed by cleaning indexedDB

PR #495, opened 2019-01-10
 Debug send transaction in prod

flexxnn:
Everything good with localhost. This issue appears only in our prod env


stcherenkov:
Reverted in #505

PR #499, opened 2019-01-10
JWWEB-35 Checking indexedDB and localStorage API in browser

Creepypoke:
Fallback page is show accordingly for this way

* Fallback page for showing inaccessible features of browser was added
* Capability checking was added


Creepypoke:
@stcherenkov 
>Move those fixes here too
Fixes already is here

PR #502, opened 2019-01-10
JWWEB-26: initial Google Analytics setup

stcherenkov:
- `ga` function that encapsulates conditional logic of sending analytics data and could also catch errors etc.
- router-specific functions to catch user transition between routes and report their anonymous data
- more specific functions for some of routes to report data with unmasked parameters

See also: [Google Analytics for Web on Confluence](https://jibrelnetwork.atlassian.net/wiki/spaces/JWALLET/pages/637239302/Google+Analytics+for+Web)


src/routes/Favorites/routes/Address/index.js
```diff
@@ -3,6 +3,10 @@
 import View from './FavoritesAddressViewContainer'
 
 export default {
-  path: 'address(/:address)',
+  path: 'address',
   component: View,
+  childRoutes: [{
+    path: ':address',
+    component: View,
+  }],
```
IViol: what's wrong with optional param `path: 'address(/:address)'`?

src/routes/DigitalAssets/routes/EditAsset/index.js
```diff
@@ -1,9 +1,16 @@
...
 import EditAssetContainer from './EditAssetContainer'
 
 const EditAsset = {
   path: 'edit-asset/:assetAddress',
+  onEnter: (nextState: ReactRouterState): void =>
+    ga(
+      'send',
+      'pageview',
+      nextState.location.pathname
+    ),
```
IViol: would be better if we have some function that hide `'send', 'pageview'` params, to avoid passing them each call.

For example
```
import { sendPageView } from 'utils/analytics'

import EditAssetContainer from './EditAssetContainer'

const EditAsset = {
  path: 'edit-asset/:address',
  onEnter: (nextState: ReactRouterState): void => sendPageView(nextState.location.pathname),
  component: EditAssetContainer,
}
```

IViol: also this allow to hide what specific service we use - `ga`, etc

src/routes/Favorites/routes/Address/index.js
stcherenkov: When user hits `.../address/0x...`, react-router fires event with final route set to `address(/:address)`. But when user hits `.../address`, react router fires the same event.

We could analyze `location.pathname` in addition to `routes`, but it looks like overkill.

src/routes/DigitalAssets/routes/EditAsset/index.js
stcherenkov: Hiding which service we use under facade is actually a bad idea.

In short, every analytics service has its own model for events / views / parameters / etc. I was designing such facade for Sberbank and we ended up with our own analytical model that mapped to every analytical service in use. This task was enormously complex and costed a lot of time.

If you would like to hear more about that, ask me in person please. :)

stcherenkov: Upgraded `ga` to `gaSendPageView` for those places

src/utils/analytics/react-router-page-view.js
```diff
@@ -0,0 +1,38 @@
...
+    prevState: ReactRouterState,
+    nextState: ReactRouterState
+  ): void => {
+    if (!exclude.find(re => re.test(nextState.location.pathname))) {
+      ga(
+        'send',
+        'pageview',
+        fullPathWithMaskedVariables(nextState)
+      )
+    }
```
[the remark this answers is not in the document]
stcherenkov: Fixed

PR #508, opened 2019-01-15
JWWEB-426 Inaccurate placeholders upon first-time wallet creation

Dragonesis:


src/components/WalletPasswordStep/WalletPasswordStep.js
```diff
@@ -82,7 +82,7 @@ class WalletPasswordStep extends PureComponent<Props> {
               errorMessage={invalidFields.passwordHint}
               color='white'
               name='password-hint'
-              placeholder='Password hint'
+              placeholder='Password hint (optional)'
```
IViol: Password hint couldn't be optional because this is the only way to recall password to unlock wallets

PR #509, opened 2019-01-15
JWWEB-441 Minor visual issues on Payment Password screen

Dragonesis:


src/components/base/JInput/JInput.js
```diff
@@ -140,7 +140,7 @@ class JInput extends PureComponent<Props> {
           <div className='lock'>
             <JIcon
               size='medium'
-              color='gray'
+              color={color}
```
IViol: there is shouldn't be lock icon on blue screens

src/components/PasswordField/Indicator/passwordFieldIndicator.scss
```diff
@@ -9,9 +9,11 @@ $password-field-indicator-height: 6px;
...
   right: $form-field-white-horizontal-padding;
   width: 60px;
-  top:
-    $form-field-padding-top +
-    (($form-field-white-font-size - $password-field-indicator-height) / 2);
+  top: 30px;
+
+  &.-offset-left {
+    right: $icon-medium-size / 1.5 + $form-field-white-horizontal-padding;
+  }
```
IViol: -offset-left but you change right attribute, looks weird

src/components/PasswordField/Indicator/passwordFieldIndicator.scss
```diff
@@ -9,9 +9,11 @@ $password-field-indicator-height: 6px;
   position: absolute;
   right: $form-field-white-horizontal-padding;
   width: 60px;
-  top:
-    $form-field-padding-top +
-    (($form-field-white-font-size - $password-field-indicator-height) / 2);
+  top: 30px;
```
IViol: we have password fields on blue and gray screens (inputs with different height), so `top` couldn't be fixed value

Dragonesis: They are the same and equal to 65px

src/components/base/JInput/JInput.js
stcherenkov: In my opinion, all except this is optional

PR #510, opened 2019-01-15
JWWEB-402 Transactions UI fixes

Dragonesis:


src/components/TransactionItem/Details/Details.js
```diff
@@ -176,14 +197,17 @@ class TransactionItemDetails extends PureComponent<Props, StateProps> {
             </div>
             <div className='value'>
               <a
+                key={to}
```
IViol: why?

src/components/TransactionItem/Details/Details.js
```diff
@@ -36,6 +36,9 @@ type Props = {|
 
 type StateProps = {|
   +isCommenting: boolean,
+  +isHoveredHash: boolean,
+  +isHoveredFrom: boolean,
+  +isHoveredTo: boolean,
 |}
```
IViol: it's worth to add just one `hovered` prop with `'hash' | 'to' | 'from' | null` type. Therefore one `onHover` handler

stcherenkov: Agreed. Also less lag while moving pointer between different items.

src/components/TransactionItem/Details/Details.js
Dragonesis: Error, remove

PR #515, opened 2019-01-17
JWWEB-289: Add X-Frame-Options to nginx config

stcherenkov:


soar:
Don't forget to check that this value won't be rewritten and will be available through all layers of balancing and proxying @jibrelnetwork/qa

soar:
Or we can set this header globally for all projects, so we don't need to change each config in each microservice

stcherenkov:
@soar not sure if we need to set this for every project.

stcherenkov:
Fixed via infrastructure

PR #516, opened 2019-01-17
upgrade & favorites minor fixes

IViol:
* Minor fixes for types of upgrade wallet
* Hide send button from favorites for readOnly wallets
* JWWEB-434 Clean favorite form on close


src/routes/Upgrade/UpgradeView.js
```diff
@@ -97,7 +97,7 @@ function UpgradeView({
 
   return (
     <CloseableScreen
-      close={isLoading ? null : onClose}
+      close={isLoading ? noop : onClose}
```
flexxnn: nooP ?????

IViol: const noop = () => undefined

PR #517, opened 2019-01-17
Fix pending txs for ETH

IViol:
**Note**: it's impossible to check that fix locally, before integrating of proxy for etherscan


src/routes/sagas/transactions.js
```diff
@@ -869,6 +861,15 @@ export function* requestTransactions(
           txs,
         ))
 
+        const txIds: TransactionId[] = Object.keys(txs)
+
+        yield all(txIds.map((txId: TransactionId) => put(transactions.checkPendingTransaction(
```
flexxnn: I think `yield all` can affect perfomance when we have really big transaction array here. I think we should find better solution here (in future)

PR #518, opened 2019-01-18
Update displaying of mint burn events

Dragonesis:


src/components/TransactionItem/Details/Details.js
```diff
@@ -32,6 +32,8 @@ type Props = {|
   +isSent: boolean,
   +isActive: boolean,
   +isFromFavorites: boolean,
+  +isEventMint: boolean,
+  +isEventBurn: boolean,
 |}
```
IViol: don't need to pass both isEventMint/Burn
it's possible to check `data.eventType === 2` that means `isMintable`

src/components/TransactionItem/Main/Main.js
```diff
@@ -117,7 +150,10 @@ class TransactionItemMain extends PureComponent<Props> {
     const amountSign: string = isSent ? '-' : '+'
     const timestamp: number = blockData.timestamp * 1000
     const isFailed: boolean = !receiptData.status || isRemoved
-    const color: TransactionTextColor = getTransactionTextColor(isSent, isFailed)
+    const color: TransactionTextColor =
+      getTransactionTextColor(isEventMint, isEventBurn, isSent, isFailed)
+    const mintMsg = isEventMint && 'Some description about mint event'
+    const burnMsg = isEventBurn && 'Some description about burn event'
```
IViol: we need to fix these strings before merging

src/services/web3.js
```diff
@@ -441,7 +441,7 @@ function checkJNTEvent(data: Object): boolean {
   )
 }
 
-function prepareJNTEvents(data: Array<Object>, assetAddress: AssetAddress): Transactions {
+function prepareJNTEvents(data: Array<Object>): Transactions {
```
IViol: but you forgot to remove passing of assetAddress from call(s) of prepareJNTEvents

PR #519, opened 2019-01-18
JWWEB-425 Integrate etherscan proxy

IViol:
**Note**: feature can't be tested locally because CORS header is not present in API server response


flow-typed/config.js
```diff
@@ -14,6 +14,9 @@ declare type JSONFormatConfig = {|
   +size: number,
 |}
 
+type BlockExplorerAPILink = 'https://api-blockchain-service.jwallet.network' |
+  'http://api-blockchain-service.jwallet.develop.stacks.jdev.network'
```
flexxnn: It is our own internal development infrastructure, I think it is not good to leave it in opensource.
Let's discuss it offline

src/config.js
```diff
@@ -107,10 +107,11 @@ const config: AppConfig = {
...
   walletIdInitTimeout: 100,
 
-  /**
-   * blockexplorer url
-   */
-  blockExplorerLink: 'etherscan.io',
+  blockExplorerUILink: 'etherscan.io',
+
+  blockExplorerAPILink: __DEV__
+    ? 'http://api-blockchain-service.jwallet.develop.stacks.jdev.network'
```
flexxnn: ^^ My comment above

flow-typed/config.js
stcherenkov: It is already in opensource. ;)

stcherenkov: I think, we should move these values to environment variables so they could be set in run-time. Like those `[MAIN_RPC_ADDR]` in index.html

stcherenkov: But in types it should be probably just string. There is no need for enum with configuration variables, it feels a little bit off...

IViol: At least for localhost dev we need default value without any params from docker image

stcherenkov: Yep, but we can pass env variables through webpack, without using Docker

stcherenkov: See here: https://github.com/jibrelnetwork/jwallet-web/blob/develop/webpack.config.js#L266-L271

PR #520, opened 2019-01-18
Correct textarea size

Dragonesis:


src/components/base/JInput/jInput.scss
```diff
@@ -164,9 +165,14 @@
       color: $dusk;
       opacity: 1;
       box-shadow: 0 3px 6px 0 rgba(61, 79, 108, 0.05);
-      min-height: $form-field-gray-height;
-      padding-left: $form-field-horizontal-padding;
-      padding-right: $form-field-horizontal-padding;
+      height: $form-field-gray-height;
```
stcherenkov: There is `rows` attribute for multiline input that has no use if we set height exactly to some px value.

We need to decide: do we have that `rows` prop or not? If we do – there is no place for fixed height of the textarea. If we don't – it should be also eliminated.

Number of rows in my opinion is more user-centric parameter than number of pixels.

Dragonesis: Ok, I'll ask Sergey

PR #523, opened 2019-01-18
JWWEB-279 Update payment password 

Creepypoke:
Work in progress, but, you are welcome to write comments and suggestions


package-lock.json
```diff
@@ -4181,10 +4181,34 @@
         "lodash": "=4.17.11"
       },
       "dependencies": {
+        "bn.js": {
```
flexxnn: One more bignumber here :)

Creepypoke: wow! I really don't know how it happen)

PR #528, opened 2019-01-21
JWWEB-448 Implement fetching of fiat courses

IViol:


env/development
```diff
@@ -1,4 +1,5 @@
 NODE_ENV=development
 BABEL_ENV=development
 GOOGLE_ANALYTICS_ID=UA-103798122-6
+TICKER_API=http://api-fiat-courses.jwallet.develop.stacks.jdev.network
```
stcherenkov: Env variable and global variable have different names. This could be misleading.

PR #531, opened 2019-01-22
Add Currency Picker Setting

stcherenkov:


src/routes/Settings/modules/settings.js
```diff
@@ -36,6 +46,11 @@ const settings = (
     case CHANGE_PAYMENT_PASSWORD:
       console.log(action, state)
```
flexxnn: fixme

stcherenkov: That's not mine. :) Will be fixed in Anton's PR. Didn't touch it to prevent merge conflicts

stcherenkov: Fixed in merge

PR #533, opened 2019-01-23
Fix view message for events burn and mint

Dragonesis:


src/components/TransactionItem/Main/Main.js
```diff
@@ -182,9 +183,19 @@ class TransactionItemMain extends PureComponent<Props> {
...
+                  value={txAddressName || txAddress}
+                  weight='bold'
+                  size='normal'
+                />
+              </div>
+            ) : (
+              <div className='address'>
+                <JText
+                  color={color}
+                  value={burnMsg || mintMsg}
```
IViol: it will raise runtime exception when burnMsg || mintMsg are both null

PR #544, opened 2019-01-24
JWWEB-436: extended google analytics setup

stcherenkov:
Added analytics middleware that catches specific state changes and reports them to Google Analytics.

[Detailed analytics description in Confluence](https://jibrelnetwork.atlassian.net/wiki/spaces/JWALLET/pages/637239302/Google+Analytics+for+Web)


src/middlewares/analyticsMiddleware/index.js
```diff
@@ -0,0 +1,19 @@
+import { newWalletEvents } from './newWalletEvents'
+import { sendTransactionEvents } from './sendTransactionEvents'
+import { userParams } from './userParams'
+
+export const analyticsMiddleware = store => next => (action) => {
```
IViol: would be better if we don't duplicate `middleware` word for any middleware.
Obvious that they are middleware, because they are placed in `middlewares` dir

src/middlewares/analyticsMiddleware/index.js
```diff
@@ -0,0 +1,19 @@
...
+  try {
+    const startState = store.getState()
+    newWalletEvents(startState, action)
+    sendTransactionEvents(startState, action)
+  } catch (err) {
```
IViol: would be better to split events by our routes (as well as reducers) "domains",
like `walletsEvents`, `digitalAssetsEvents` (word `Events` better to skip also)

src/middlewares/analyticsMiddleware/userParams.js
```diff
@@ -0,0 +1,88 @@
...
+    case SET_WALLETS:
+    case SET_WALLETS_ITEMS: {
+      /* eslint-disable no-param-reassign, fp/no-mutation */
+      // because reduce is designed to work this way
+      const stats = selectWalletsItems(state).reduce((memo, wallet) => {
+        memo.total += 1
+        if (wallet.isReadOnly) {
+          memo.readonly += 1
+        }
+        if (wallet.type === 'mnemonic') {
```
IViol: we have `checkMnemonicType` util

src/middlewares/analyticsMiddleware/newWalletEvents.js
```diff
@@ -0,0 +1,51 @@
+import { gaSendEvent } from 'utils/analytics'
```
Creepypoke: I guess, it's `walletCreateEvents` name is better.

stcherenkov: There are both create and import events here

PR #545, opened 2019-01-25
Add modernizr check: crypto.getRandomValues inside web workers

flexxnn:
This check will reject all browsers with unsupported crypto.getRandomValues inside web workers
Now it is IE and Microsoft edge


src/browsercheck.js
```diff
@@ -10,20 +10,23 @@ const browsercheck: () => Promise<void> = () => {
...
+        !workerscryptoDBAvailable
+        ) {
+          return reject(
+            new Error('Some of required browser APIs are not available')
+          )
+        }
 
-      return resolve()
+        return resolve()
+      })
```
IViol: to make these checks clean we need to resolve Promise for each of them, like
```
return checkFaturesSync
  .then(checkIndexedDB)
  .then(checkWorkerRNG)
```

stcherenkov: Better to have helper function that gets list of async tests and fires once all of them are checked. Like:

```
function checkAll (testNames, callback) {
  Promise.all(
    testNames.map((name) =>
      new Promise((resolve, reject) => 
        Modernizr.on(name, (result) => result ? resolve() : reject())
      )
    )
  )
}
```

And call it like:
```
checkAll(['indexeddb', 'workerscrypto'])
  .then(...)
```

In this case, when you add another async check, you need to add only its name to the list. This also means we could pass list of checks via configuration.

But please don't fix it right now, it's fine this way too.

PR #550, opened 2019-01-28
Transaction fail warning

flexxnn:


src/routes/DigitalAssets/routes/Send/components/Confirm/Card/Card.js
```diff
@@ -44,6 +46,14 @@ function DigitalAssetsSendConfirmCard({
   return (
     <div className='digital-assets-send-confirm-card'>
       <div className='content'>
+        {notifyPotentiallyFail &&
+        <div className='willfail'>
+          <JText
+            value='Your transaction will potentially fail'
```
stcherenkov: "Your transaction does not pass one or more check-ups, so we assume **it will most likely fail**. Please make sure you know what you are doing before proceeding"

PR #553, opened 2019-01-28
JWWEB-319: disclaimer screen

Dragonesis:


src/components/base/JCheckbox/JCheckbox.js
```diff
@@ -43,13 +53,14 @@ class JCheckbox extends PureComponent<Props> {
           <span className='flag' />
           <span className='label'>
             <JText
-              color='gray'
+              color={color}
               size='normal'
               value={label}
-              weight='bold'
+              weight={isRegular ? 'regular' : 'bold'}
```
IViol: just pass `null`, or make this prop required, because now there are two ways to define "regular" font weight

src/components/base/JCheckbox/JCheckbox.js
```diff
@@ -43,13 +53,14 @@ class JCheckbox extends PureComponent<Props> {
...
               whiteSpace='wrap'
             />
           </span>
+          { children }
```
IViol: spaces are useless in jsx

src/components/base/JText/JText.js
```diff
@@ -6,7 +6,7 @@ import React, { PureComponent } from 'react'
 type JTextAlign = 'center'
 type JTextFontCase = 'upper'
 type JTextDecoration = 'underline'
-type JTextWeight = 'bold' | 'bolder'
+type JTextWeight = 'regular' | 'bold' | 'bolder'
```
IViol: need to finish it or remove

src/routes/Agreements/AgreementsView.js
```diff
@@ -0,0 +1,116 @@
...
+            <JText
+              size='title'
+              color='white'
+              value='Terms and Conditions'
+              whiteSpace='wrap'
+              align='center'
+            />
+          </h1>
+          <div className='items'>
+            {conditionsArray.map((item: string, i) => (
```
IViol: need to use some ids, by condition name I asume
and avoid using of array indexes as key attr

src/utils/agreements/agreements.js
```diff
@@ -0,0 +1,43 @@
+const agreements = {}
```
IViol: please create separate file for each util function

flow-typed/redux/agreements.js
```diff
@@ -0,0 +1,3 @@
+// @flow
+
+declare type NameConditionsType = 'condition-0' | 'condition-1' | 'condition-2' | 'condition-3'
```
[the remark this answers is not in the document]
stcherenkov: I.e.:
- 'understandPrivateDataPolicy'
- 'consentNoWarranty'
- 'consentTrackingCookies'
- 'acceptTermsAndConditions'

src/routes/Agreements/AgreementsView.js
```diff
@@ -0,0 +1,116 @@
...
+
+import React, { PureComponent } from 'react'
+import { JCheckbox, JText, JRaisedButton } from 'components/base'
+import { setAgreements, getAgreements, checkAgreements } from 'utils/agreements'
+
+type StateProps = {|
+  +isDisabled: boolean,
+|}
+
+const conditionsArray = [
```
stcherenkov: It would be better to assign texts as key-value pairs. So when we will change something, we do not need to support the same order in multiple places.

src/utils/agreements/agreements.js
```diff
@@ -0,0 +1,43 @@
...
+
+/* eslint-disable fp/no-mutation */
+export const getAgreements = () => JSON.parse(localStorage.getItem('agreements'))
```
stcherenkov: Nope, each agreement should be on separate localStorage key.

src/routes/Agreements/AgreementsView.js
stcherenkov: Why? Because we will check and modify them separately. So we could discard `JSON.parse` which is a little too resource-consuming.

src/utils/agreements/agreements.js
stcherenkov: Why? Because we will check and modify them separately. So we could discard `JSON.parse` which is a little too resource-consuming.

src/utils/agreements/checkAgreements.js
```diff
@@ -0,0 +1,16 @@
+import getAgreements from './getAgreements'
+
+export default function checkAgreements(nameConditionsArray) {
+  const check = nameConditionsArray.map((key) => {
```
stcherenkov:
```
return nameConditionsArray.every((key) =>
  getAgreements(key) === 'true'
)
```

src/utils/agreements/getAgreements.js
```diff
@@ -0,0 +1,3 @@
+export default function getAgreements(agreement) {
```
stcherenkov: It gets only one agreement check value, so better to call it in singular.

src/utils/agreements/getAgreements.js
```diff
@@ -0,0 +1,3 @@
...
+  return localStorage.getItem(agreement) === null ? 'false' : localStorage.getItem(agreement)
```
stcherenkov: Actually we never need string representation for this value, so it is better to write it as:
```
return localStorage.getItem(agreement) === 'true'
```

src/routes/Agreements/AgreementsView.js
```diff
@@ -54,48 +59,49 @@ class AgreementsView extends PureComponent<Props, StateProps> {
...
-              <div className='item' key={item}>
-                <JCheckbox
-                  onChange={this.onChange(`condition-${i}`)}
-                  label={item}
-                  color='white'
-                  name={`conditions-${i}`}
-                  isChecked={agreementsFlags[`condition-${i}`]}
-                  isRegular
-                />
+            {Object.keys(conditionsArray).map((key: string) => (
```
stcherenkov: `Object.keys(conditionsArray)` does not guarantee elements order.

For ordered items you already have `nameConditionsArray`. Why did you сhange it?

src/routes/Agreements/AgreementsView.js
```diff
@@ -54,52 +63,53 @@ class AgreementsView extends PureComponent<Props, StateProps> {
...
+                    name={key}
+                    isChecked={getAgreementValue(key)}
+                    isRegular
+                  />
+                ) : (
+                  <JCheckbox
+                    onChange={this.onChange(key, !getAgreementValue(key))}
+                    color='white'
+                    label='I have read and accepted'
+                    name='conditions-3'
```
IViol: name='acceptTermsAndConditions'

stcherenkov: ouch

stcherenkov: fixed

src/routes/Agreements/AgreementsViewContainer.js
```diff
@@ -1,9 +1,16 @@
...
 
 import { connect } from 'react-redux'
+import { push } from 'react-router-redux'
 
 import AgreementsView from './AgreementsView'
 
+const mapStateToProps = () => ({})
+
+const mapDispatchToProps = {
+  onSubmit: () => push('/wallets/start'),
```
IViol: I think better push `back` with fallback `/`

src/utils/agreements/setAgreementValue.js
```diff
@@ -0,0 +1,9 @@
+export default function setAgreementValue(agreement, value) {
+  try {
+    localStorage.setItem(agreement, value)
+  } catch (e) {
+    if (e === 'QUOTA_EXCEEDED_ERR') {
+      console.error('Limit exceeded')
+    }
```
IViol: just `console.error`?

src/routes/Agreements/AgreementsViewContainer.js
stcherenkov: fixed

src/utils/agreements/setAgreementValue.js
stcherenkov: fixed

PR #559, opened 2019-01-28
JWWEB-437: Routing

stcherenkov:
The main idea was to encapsulate all the routing rules in a single file, so future refactor would be more simple.

There is also a hack suggested by @IViol (thnx again!) with redirect happening after all redux-persist hydration events.


src/middlewares/redirect/filters/agreement.js
```diff
@@ -1 +1,17 @@
...
+// @flow
+
+import { checkAgreements } from 'utils/agreements'
+import { CONDITIONS_LIST } from 'data/agreements'
+
+export const agreement = (state: AppState, pathname: string): string => {
+  const isAgreementsPathname = /^\/agreements\/?$/.test(pathname)
+  const isAllAgreementsChecked = checkAgreements(CONDITIONS_LIST)
+  if (isAgreementsPathname && isAllAgreementsChecked) {
+    return '/wallets'
```
IViol: return '/'

stcherenkov: fixed

PR #587, opened 2019-02-01
JWWEB-527 Delete wallet - wrong redirect if user pressing "nope, stop it"

Creepypoke:
Routing on the delete wallet page was refactored


src/components/ButtonWithConfirm/ButtonWithConfirm.js
```diff
@@ -81,6 +82,7 @@ class ButtonWithConfirm extends Component<Props, ComponentState> {
   resetCountdown = () => {
     this.finishCountdown()
     this.setCountdown(this.props.confirmTimeout)
+    this.props.onCancelClick()
```
IViol: need to check that `onCancelClick` was passed

PR #596, opened 2019-02-04
Input with icon

Dragonesis:
JWWEB-537
Cross is superimposed on the text in the Comment field


src/components/base/JInput/JInput.js
```diff
@@ -26,6 +26,7 @@ type Props = {|
   +color: JInputColor,
   +type: JInputType,
   +sideBorderRadius: 'all' | 'top' | 'left' | 'bottom' | 'right',
+  +iconPosition: 'left' | 'right' | null,
```
stcherenkov: What does icon position = null mean?

PR #597, opened 2019-02-04
JWWEB-519 Resyncing of transactions event was added to google analytics

Creepypoke:


src/middlewares/analyticsMiddleware/index.js
```diff
@@ -1,12 +1,16 @@
 import { newWalletEvents } from './newWalletEvents'
 import { sendTransactionEvents } from './sendTransactionEvents'
 import { userParams } from './userParams'
+import { resyncTransactionsEvents } from './resyncTransactionsEvents'
 
+// Declare here all Google analytics events:
+// @see https://jibrelnetwork.atlassian.net/wiki/x/BoD7JQ
```
IViol: Better to avoid using of such links in comments, because they often lose actuality.

PR #599, opened 2019-02-04
JWWEB-470 Search results are displayed after back from transactions through digital asset button

Creepypoke:
JSearch clears value on unmount


src/components/base/JSearch/JSearch.js
```diff
@@ -39,6 +39,10 @@ class JSearch extends PureComponent<Props, ComponentState> {
     }
   }
 
+  componentWillUnmount() {
+    this.onChange('')
```
stcherenkov: Add FIXME comment here plz

Looks like it needs further investigation in the future

PR #604, opened 2019-02-04
JWWEB-493: Update for token icons

stcherenkov:


src/components/base/JAssetSymbol/jAssetSymbol.scss
```diff
@@ -45,51 +72,80 @@
...
     &.-symbol-xaur { background-image: $token-xaur-blue; }
+    &.-symbol-xin { background-image: $token-xin-blue; }
+    &.-symbol-xpa { background-image: $token-xpa-blue; }
+    &.-symbol-zil { background-image: $token-zil-blue; }
     &.-symbol-zrx { background-image: $token-zrx-blue; }
   }
 
   &.-gray {
-    background-image: linear-gradient(to top, $dusk, $dusk);
+    &.-symbol-not-listed { color: $blue; }
```
IViol: `{ color: $dusk; }` ?

src/components/base/JAssetSymbol/symbolsAvailable.js
```diff
@@ -0,0 +1,102 @@
+export const SYMBOLS_AVAILABLE = [
```
IViol: would be better to place it in `src/data`

src/components/base/JAssetSymbol/jAssetSymbol.scss
stcherenkov: Gray text is invisible on white background :(

src/components/base/JAssetSymbol/symbolsAvailable.js
stcherenkov: It is not global data actually, but view parameters required only for symbol rendering. So they are context to their domain now.

It would be even better not to store it at all, but to get it automatically from file system.

But what do we say to such tasks? Not today!

PR #610, opened 2019-02-05
JWWEB-554: Wallet has its' icon back

stcherenkov:


src/components/MenuPanel/WalletManager/Main/Main.js
```diff
@@ -18,7 +18,6 @@ type Props = {|
   +address: string,
   +isActive: boolean,
   +isReadOnly: boolean,
-  +isSimplified: boolean,
```
IViol: please remove this prop from passing to this component

stcherenkov: Fixed

PR #634, opened 2019-02-07
JWWEB-566: incorrect sorting

stcherenkov:
Now we take into account number of decimals when we sort by balance

### DIFF
--- a/src/utils/digitalAssets/compareDigitalAssetsByBalance.js
+++ b/src/utils/digitalAssets/compareDigitalAssetsByBalance.js
@@ -1,18 +1,27 @@
 // @flow
 
-import toBigNumber from 'utils/numbers/toBigNumber'
+import { divDecimals, toBigNumber } from 'utils/numbers'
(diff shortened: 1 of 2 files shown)


src/utils/digitalAssets/compareDigitalAssetsByBalance.js
```diff
@@ -1,18 +1,27 @@
...
+  const { decimals } = asset.blockchainParams
+  if (!balance || !decimals) {
+    return toBigNumber()
+  }
+
+  return divDecimals(
+    toBigNumber(balance.value),
+    decimals
+  )
+}
```
IViol:
```
return divDecimals(
    toBigNumber(balance.value),
    decimals
  )
```
could be just `return divDecimals(balance.value, decimals)`

stcherenkov: Thnx, fixed

PR #647, opened 2019-02-13
Fix regexp searching

IViol:


src/utils/transactions/searchTransactions.js
```diff
@@ -1,33 +1,42 @@
...
   +fromName: ?string,
 }
 
+const HEX_PREFIX: string = '0x'
+
 function searchBy(f: Function, query: string): Function {
   return (value: ?string): boolean => value ? f(value, query) : false
 }
 
 function checkHashes(field: string, searchQuery: string): boolean {
```
stcherenkov: A good place for a unit test, btw

PR #659, opened 2019-02-14
JWWEB-591 Unable to send first transaction with opened nonce field

flexxnn:


src/routes/DigitalAssets/routes/Send/sagas/digitalAssetsSend.js
```diff
@@ -377,13 +377,13 @@ function* checkNonce(formFieldValues: DigitalAssetsSendFormFields): Saga<void> {
 
   const nonce: ?number = yield* requestNonce('latest')
 
-  if (!nonce) {
+  if (nonce == null) { // null or undefined, but not zero
```
stcherenkov: Check out `/src/utils/type/isVoid.js`?

PR #666, opened 2019-02-15
JWWEB-605 Icons component that manages and provides icons by name

Dragonesis:
Displayed icons in the livebook
Implemented sprite for assets.
Fixed some obvious conflicts.

Icons can respond to hover, also accept props. Not sure yet of the need to use the currentColor property


stcherenkov:
Huge +1 for sizes as numbers

src/components/base/JIcon/JIcon.js
```diff
@@ -9,19 +10,26 @@ type Props = {
   name: string,
   size: JIconSize,
   color: JIconColor,
+  isCurrentColor: boolean,
 }
 
+const files = require.context('../../../public/assets/icons/new-pack', true, /.*\.svg$/)
+files.keys().forEach(x => files(x))
```
stcherenkov: Looks like black magic is done here. Could you please describe what is happening?

webpack.config.js
```diff
@@ -130,6 +130,40 @@ module.exports = {
         // match the requirements. When no loader matches it will fall
         // back to the "file" loader at the end of the loader list.
         oneOf: [
```
stcherenkov: Looks like we can achieve multiple sprites through multiple configurations with different `include` paths and [named files extraction](https://github.com/kisenka/svg-sprite-loader#sprite-filename).

But I'm not sure. Anyway, we don't need it right now – it's just one of possible solutions for our future nonexistent problem. 😉

stcherenkov:
Great work, keep it up!

src/components/base/JIcon/JIcon.js
```diff
@@ -2,36 +2,47 @@
...
+  }),
+  'id',
+)
 
 class JIcon extends PureComponent<Props> {
   static defaultProps = {
-    size: 'medium',
     color: null,
-    useСurrentСolor: false,
+    useFill: false,
```
stcherenkov: How do I know from outside of the icon if it uses fill or not? Looks like this depends only on icon id too.

Maybe we could skip guessing and just set up both fill and stroke color for icon? If it does not have stroke, it would have stroke-width set to 0 inside of its svg source.

Dragonesis: So far, I have no idea how to check it. We previously remove the fill and stroke attributes from the icons at the assembly stage. Now there are only two such icons.

stcherenkov:
Also please revert assets submodule to commit 84df030f214d67be8fc1284682a128c1e1c3373c. We are still not sure about those latest changes.

src/components/base/JIcon/JIcon.js
```diff
@@ -1,33 +1,53 @@
...
-      name,
-      size,
-      color,
-    }: Props = this.props
-
-    return (
-      <div className={`j-icon -${name} -${size} -${color}`} />
-    )
-  }
+const files = require.context('../../../public/assets/icons/sprite-pack', true, /.*\.svg$/)
```
Creepypoke: Maybe, we can to move this "dirty" part of code to other file and call them like Hight-Order Function:
```javascript
import JIcon from '...'
function importIconsFiles (props: Props) {
  /*
    dirty part with imports files
  */

  return <JIcon {...props} />
}
```

Creepypoke: Maybe we can move this part to other file and call here?

src/components/base/JAssetSymbol/JAssetSymbol.js
```diff
@@ -14,11 +15,32 @@ type Props = {
...
-)
+const files = require.context('../../../public/assets/tokens/blue', true, /.*\.svg$/)
+const icons = files.keys().map(x => files(x).default).reduce((result, {
+  id, url,
+}) => ({
+  ...result,
+  [id]: url,
+}), {})
+
+class JAssetSymbol extends PureComponent<Props> {
```
Creepypoke: From my point of view, we need to using `function declaration` for stateless and methodless components

src/components/base/JIcon/JIcon.js
```diff
@@ -27,29 +26,28 @@ const icons = keyBy(
...
-      >
-        <use xlinkHref={iconData.url} />
-      </svg>
-    )
-  }
+function JIcon({
+  name, color,
+}: Props) {
+  const iconData = icons[`${name}-usage`]
+  const isUseFill = name.indexOf('use-fill') !== -1
```
stcherenkov: `hasFill`?

.storybook/webpack.config.js
```diff
@@ -51,12 +51,53 @@ module.exports = (baseConfig, env) => {
...
       name: 'static/media/[name].[hash:8].[ext]',
     }
   })
 
+  baseConfig.module.rules.push({
+    test: /\.svg$/,
+    include: [
+      path.resolve(srcPath, 'public/assets/icons/sprite-pack'),
+      path.resolve(srcPath, 'public/assets/tokens'),
+    ],
```
IViol: 1)
```
    exclude: [
      path.resolve(srcPath, 'public/assets/icons/sprite-pack'),
      path.resolve(srcPath, 'public/assets/tokens'),
    ],
```
2)
```
    include: [
      path.resolve(srcPath, 'public/assets/icons/sprite-pack'),
      path.resolve(srcPath, 'public/assets/tokens'),
    ],
```
would be nice to reuse this list

webpack.config.js
```diff
@@ -124,7 +126,42 @@ module.exports = {
         ],
         include: PATHS.SOURCE,
       },
-
+      {
+        test: /\.svg$/,
+        include: [
+          path.resolve(__dirname, 'src/public/assets/icons/sprite-pack'),
+          path.resolve(__dirname, 'src/public/assets/tokens'),
```
IViol: the same comment about reusing of this list

PR #675, opened 2019-02-19
Fix storybook

flexxnn:


stories/index.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 
 import '../src/styles/core.scss'
-import './styles.css'
```
IViol: `styles.css` is importing from another place?

.storybook/webpack.config.js
```diff
@@ -24,7 +57,21 @@ module.exports = (baseConfig) => {
...
 
-  baseConfig.resolve.modules = [srcPath, 'node_modules']
+  baseConfig.plugins.push(
+    new webpack.DefinePlugin({
+      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
+      'process.env.BUILD_NUMBER': 'dev',
+      '__DEV__': true,
+      '__PROD__': false,
+      '__DEFAULT_TICKER_API__': '{}',
+      '__DEFAULT_BLOCKEXPLORER_API__': '{}',
```
IViol: why not just empty strings?

.storybook/webpack.config.js
```diff
@@ -1,21 +1,54 @@
...
+        loader: require.resolve('sass-loader'),
+        options: {
+          includePaths: [
+            path.resolve('src'),
+          ],
+        },
       },
-      'sass-loader'
     ],
   })
```
IViol: isn't it possible to reuse these configs for loaders in both webpack configs?

stcherenkov: It looks like reusing requires more effort (and reduces webpack.config maintainability) than just copy-pasting on demand.

.storybook/webpack.config.js
```diff
@@ -24,7 +57,21 @@ module.exports = (baseConfig) => {
     }
   })
 
-  baseConfig.resolve.modules = [srcPath, 'node_modules']
+  baseConfig.plugins.push(
+    new webpack.DefinePlugin({
```
stcherenkov: If we use only `DefinePlugin` from webpack, maybe we should destructure `require('webpack')`? That's not necessary, but it may improve readability.

PR #676, opened 2019-02-20
JWWEB-633: Add CSS modules support with small usage example

stcherenkov:
Right now it is configured under `.m.scss`.

Good:
- CSS file is a lot more flat
- there is no name collisions, guaranteed
- we can remove unused container classes that were used only for handling nesting properly
- writing this list took more than single file migration, so we could probably migrate the whole code base in a day

Bad:
- camelCase coercion may be obscure for those new to CSS modules (can be turned off though)

Let's discuss!


src/routes/Agreements/AgreementsView.js
```diff
@@ -55,9 +57,9 @@ class AgreementsView extends PureComponent<Props, StateProps> {
               align='center'
             />
           </h1>
-          <div className='items'>
+          <div>
             {CONDITIONS_LIST.map((key: string) => (
-              <div className='item' key={key}>
+              <div className={style.checkboxItem} key={key}>
```
stcherenkov: This name is used purely as example of camelCase coercion. It would be a lot more simple to call it just `item`.

flexxnn:
👍

flexxnn:
Do we need to fix stylelint config to work with CSS modules?

stcherenkov:
@flexxnn It does not show any issues right now, but I'll double check

stcherenkov:
> @flexxnn It does not show any issues right now, but I'll double check

Checked, everything looks fine.

PR #685, opened 2019-02-27
Fix issue with sagas names conflict

IViol:


src/store/configureStore.js
```diff
@@ -61,11 +56,7 @@ function configureStore(initialState: $Shape<AppState> = {}, history: Object) {
   // ======================================================
   // Run sagas
   // ======================================================
-  forOwn(sagas, saga =>
-    isObject(saga) ?
-      forOwn(saga, sagaMiddleware.run)
-      : sagaMiddleware.run(saga),
-  )
+  Object.keys(sagas).forEach(sagaName => sagaMiddleware.run(sagas[sagaName]))
```
Creepypoke: For this case you can just using `forOwn(sagas, sagaMiddleware.run)`

PR #686, opened 2019-02-27
Fix delete wallet timeout for dev env

IViol:


src/config.js
```diff
@@ -132,7 +132,7 @@ const config: AppConfig = {
   /**
    * timeout before confirming of delete wallet action
    */
-  deleteConfirmTimeout: 30,
+  deleteConfirmTimeout: __DEV__ ? 0 : 30,
```
stcherenkov: Let's make it at least 3 sec to have a tiny chance to spot a bug if it happens?

PR #690, opened 2019-03-04
Feature/jsx css rework

Creepypoke:


src/components/base/JRaisedButton/JRaisedButton.js
```diff
@@ -1,119 +1,75 @@
...
-      <div
-        onMouseEnter={handle(this.onHover)(true)}
-        onMouseLeave={handle(this.onHover)(false)}
-        onClick={isDisabled ? undefined : onClick}
+      <button
+        type='button'
+        {...omittedProps}
         className={buttonClassName}
+        onClick={this.props.onClick}
+        disabled={isDisabled}
```
IViol: need to add `disabled` to `omittedProps`

src/components/base/JRaisedButton/JRaisedButton.js
```diff
@@ -1,119 +1,75 @@
...
-  +isLoading: boolean,
-  +isDisabled: boolean,
-|}
+type Props = {
+  children: React$Node,
+  theme: Themes,
+  onClick?: Function,
+  isLoading?: boolean,
+  disabled?: boolean,
+  className?: ?string,
```
flexxnn: Should we use `className: string | Object` to pass some Object values to cx too?

src/routes/Upgrade/UpgradeView.js
```diff
@@ -53,6 +51,9 @@ const renderPasswordField = (isInvalidPassword: boolean) => (fieldProps) => {
   return <JInputField {...fieldProps} />
 }
 
+/**
+ * @return {null}
+ */
```
IViol: ?

src/components/base/JRaisedButton/JRaisedButton.js
Creepypoke: I guess, we should using cx explicitly in context.

PR #700, opened 2019-03-14
JWWEB-620: Initial router5 setup

stcherenkov:
# Core concepts

1. Routing state machine happens before business-logic, so routing is independent layer
2. Description of routing states and ability to move from one to another is gathered in 1 file so it could be observed all at once
3. There is difference between path and name. Path is public interface for browser (or user) to tell us they want to move state machine into specific state. Name is internal API that tells how to move throw state machine and what part of it we should check out
4. Routing layer has 2 sets of API – redux-like and react-like, redux-like takes actions, react-like requires links (see JLink as try to simplify access to that API)


src/components/base/JLink/JLink.js
```diff
@@ -0,0 +1,54 @@
...
+  const { href } = initialProps
+  const props = omit(initialProps, [
+    'theme',
+    'className',
+    'activeClassName',
+    'rel',
+  ])
+  const className = classnames(
+    initialProps.theme && jLinkStyle[initialProps.theme],
+    initialProps.className,
```
IViol: would be great to destructure `theme` & `className` props as well as `href`

src/components/base/JLink/tests/JLink.test.js
```diff
@@ -0,0 +1,93 @@
...
+    )
+
+    expect(wrapper.text()).toBe('some path')
+    expect(wrapper.find('a').length).toBe(1)
+    expect(wrapper.find('a').prop('href')).toBe('/some/path')
+    expect(wrapper.find('a').prop('target')).toBeUndefined()
+  })
+
+  test('calls router on internal link click', () => {
+    const router = createRouter5([
```
IViol: it's ok to reuse it

src/components/base/JLink/JLink.js
stcherenkov: Fixed

src/components/base/JLink/tests/JLink.test.js
stcherenkov: Router itself needs to be created each time to guarantee tests independence. But config can be reused – done.

src/routes/NotFound/NotFoundViewContainer.js
```diff
@@ -2,14 +2,8 @@
...
-
 import NotFoundView from './NotFoundView'
 
-const mapDispatchToProps = {
-  goToHome,
-}
-
 export default (
-  connect/* :: < AppState, any, OwnPropsEmpty, _, _ > */(null, mapDispatchToProps)
+  connect/* :: < AppState, any, OwnPropsEmpty, _, _ > */(null, null)
```
IViol: so, `connect` could be removed

stcherenkov: fixed

flow-typed/redux/appState.js
```diff
@@ -7,6 +7,19 @@ import {
   type Channel,
 } from 'redux-saga'
 
+declare type Router5Route = {
+  +name: string,
+  +params: {
+    [key: string]: any,
+  },
+  +path: string,
+}
```
Creepypoke: Please, add `TODO` for `meta` property pass

src/components/AssetItem/AssetItem.js
```diff
@@ -15,17 +15,10 @@ import {
...
 
 import {
-  JCard,
-  JIcon,
-  JText,
-  JSwitch,
-  JTooltip,
-  JAssetSymbol,
-  JFlatButton,
+  JCard, JIcon, JText, JSwitch, JTooltip, JAssetSymbol, JFlatButton, JLink,
```
Creepypoke: ESlint error

src/components/base/JLink/JLink.js
```diff
@@ -0,0 +1,76 @@
...
+  checkIsExternal,
+} from 'utils/uri'
+
+import jLinkStyle from './JLink.m.scss'
+import { JLinkInternal } from './JLinkInternal'
+
+export type Theme = 'text-white' | 'text-blue'
+
+// base component with inexact props
+export type JLinkProps = {
```
Creepypoke: Looks like typing should be `StyledComponent<Theme> & { ... }`?

stcherenkov: Does not work for me, I need this component to be available completely without styles (meaning – without theme).

src/components/AssetItem/AssetItem.js
stcherenkov: Fixed in related PR #719

flow-typed/redux/appState.js
stcherenkov: Fixed in related PR #719

PR #701, opened 2019-03-18
JWWEB-607 Removing handling hover via javascript

Creepypoke:


src/components/AssetItem/AssetItem.js
```diff
@@ -38,8 +38,7 @@ type Props = {|
 
 type StateProps = {|
   +isToggled: boolean,
-  +isHoveredEdit: boolean,
-  +isHoveredTrash: boolean,
+  +showDelete: boolean,
```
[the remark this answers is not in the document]
Creepypoke: Completely agree

src/components/AssetItem/AssetItem.js
```diff
@@ -135,33 +129,44 @@ class AssetItem extends PureComponent<Props, StateProps> {
...
                     />
                   </JTooltip>
                 </div>
-                <div
-                  className='item -delete'
-                  onMouseEnter={this.handleMouseEnterRemove}
-                  onMouseLeave={this.handleMouseLeaveRemove}
-                  onClick={this.handleMouseLeaveRemove}
-                >
-                  <ButtonWithConfirm
```
IViol: why you decided to remove ButtonWithConfirm and place its code here?

src/components/FavoriteItem/FavoriteItem.js
```diff
@@ -110,48 +106,52 @@ class FavoriteItem extends PureComponent<Props, ComponentState> {
...
-                bgColor='white'
-                labelCancel={t`No`}
-                iconTooltipName='trash'
-                labelConfirm={t`Yes, delete`}
-                isReverse
-              />
+            <div className='item -delete'>
+              <div className='confirms'>
+                {
+                  this.state.showDelete ?
```
IViol: copy-paste the same code in several places instead of re-using ButtonWithConfirm component?

src/components/base/JFlatButton/JFlatButton.js
```diff
@@ -93,25 +75,24 @@ class JFlatButton extends PureComponent<Props, StateProps> {
       isDisabled,
       isBordered,
       isTransparent,
-      isUnderscored,
```
IViol: is it unused?

src/components/FavoriteItem/FavoriteItem.js
Creepypoke: It's not same scenarios. Because all cases (3) of using this component is different way, and, in my opinion, more straightforward design is: when we make three different implementation.

src/components/AssetItem/AssetItem.js
Creepypoke: I think we should to remove ButtonWIthConfirm from codebase.

src/components/base/JFlatButton/JFlatButton.js
Creepypoke: Looks like yes

src/components/FavoriteItem/FavoriteItem.js
```diff
@@ -24,24 +24,24 @@ type Props = {|
   +isWalletReadOnly: boolean,
 |}
 
-type FavoriteItemHoveredIcon = 'send' | 'edit' | 'trash'
-
 type ComponentState = {|
-  +hovered: ?FavoriteItemHoveredIcon,
   +isActionsToggled: boolean,
+  +showDelete: boolean,
```
IViol: here naming should be changed as well

Creepypoke: Reasonable, my bad :(

PR #702, opened 2019-03-19
JWWEB-638 Update dependences

flexxnn:


package.json
```diff
@@ -112,51 +109,51 @@
...
+    "postcss-flexbugs-fixes": "4.1.0",
     "postcss-loader": "3.0.0",
-    "react-app-polyfill": "0.1.3",
-    "react-dev-utils": "6.1.1",
+    "react-app-polyfill": "0.2.2",
+    "react-dev-utils": "8.0.0",
     "react-error-overlay": "1.0.10",
-    "react-test-renderer": "15.6.2",
+    "react-test-renderer": "16.8.4",
     "recompose": "0.30.0",
```
IViol: do we need `recompose`?

package.json
```diff
@@ -36,55 +37,51 @@
...
-    "dotenv": "4.0.0",
+    "classnames": "2.2.6",
+    "dotenv": "7.0.0",
     "events": "3.0.0",
-    "file-saver": "1.3.3",
-    "final-form": "4.11.0",
+    "file-saver": "2.0.1",
+    "final-form": "4.12.0",
     "history": "3.3.0",
     "js-sha3": "0.8.0",
```
IViol: `js-sha3`? do we use it?

package.json
flexxnn: Recompose in the devDependences, it can be used with storybook

package.json
flexxnn: Good point, fixed.

PR #707, opened 2019-03-22
JWWEB-697 JTextArea

Creepypoke:


src/components/base/JTextArea/JTextArea.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function handleChange(finalFormHandler: Function): SyntheticEvent<HTMLTextAreaElement> => void {
+  return function textAreaChangeHandler(fieldEvent) {
+    heightCalc(fieldEvent)
+    finalFormHandler(fieldEvent)
+  }
+}
+
+export class JTextArea extends PureComponent<Props, void> {
```
stcherenkov: This component is too specific, to be named `JTextArea` and to come under `base` namespace. Consider moving it and renaming according to its usage context.

src/components/base/JTextArea/JTextArea.js
```diff
@@ -0,0 +1,67 @@
...
+
+    return (
+      <textarea
+        {...omitedProps}
+        className={classNames(
+          this.props.className,
+          jTextAreaStyle.core,
+          jTextAreaStyle[this.props.theme],
+        )}
+        onChange={handleChange(this.props.onChange)}
```
stcherenkov: This requires html rerender every time, even when props do not change.

One path is to declare `handleChange` as class method, like this:
```
handleChange = () => {
  const value = ...
  this.props.onChange(value)
}
```

Another path is to memoize function creation somehow.

Your choice.

src/components/base/JTextArea/JTextAreaField.js
```diff
@@ -0,0 +1,27 @@
...
+  ...HTMLTextAreaElement,
+}
+
+export function JTextAreaField(props: Props) {
+  const message = props.meta.visited ? String(props.meta.error) : ''
+
+  return (
+    <div>
+      <JTextArea
+        {...props}
```
stcherenkov: Maybe omit `input` from `props`? Or maybe pass only `props.input`?

src/components/base/JTextArea/JTextAreaField.js
```diff
@@ -0,0 +1,27 @@
...
+
+import { JTextArea } from './JTextArea'
+
+type FieldRenderProps = ExtractReturn<typeof Field>
+type Props = FieldRenderProps & {
+  ...HTMLTextAreaElement,
+}
+
+export function JTextAreaField(props: Props) {
+  const message = props.meta.visited ? String(props.meta.error) : ''
```
stcherenkov: I suggest we add this check as utility function, because it is typical form behaviour and we could need to change it all at once.

Creepypoke: Completely agree, It's more flexible way

flow-typed/components.js
```diff
@@ -10,3 +10,8 @@ declare type AddressPickerAddress = {|
   address: Address,
   title: string,
 |}
+
+declare type StyleComponent<Theme: string> = {
+  theme: Theme,
```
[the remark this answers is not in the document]
Creepypoke: Sounds reasonable, but what's mean [Style component](https://github.com/jibrelnetwork/frontend/blob/master/architecture/components.md#style-components):
> This component must accept `theme` property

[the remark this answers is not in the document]
stcherenkov: Yep, `Maybe string` is the best call

src/components/base/ErrorMessage/ErrorMessage.js
```diff
@@ -0,0 +1,34 @@
...
+      className,
+      ErrorMessageStyle.core,
+      ErrorMessageStyle[theme])}
+    >
+      {message}
+    </span>
+  )
+}
+
+const MemorizedErrorMessage = React.memo/* :: <Props> */(ErrorMessage)
```
IViol: `Memoized` is correct [term](https://en.wikipedia.org/wiki/Memoization#Etymology)

src/components/base/JTextArea/JTextAreaField.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+
+import React, { Fragment } from 'react'
+import { type FieldRenderProps /* FieldRenderProps */ } from 'react-final-form/dist/types.js.flow'
```
IViol: is it necessary to specify `.js.flow`?

src/components/base/ErrorMessage/ErrorMessage.js
```diff
@@ -0,0 +1,34 @@
...
+type Props = StyleComponent<Theme> & {
+  message: string,
+}
+
+function ErrorMessage(
+  {
+    theme = 'error',
+    message = '',
+    className,
+  }: Props) {
```
IViol: it could be written as
```
function ErrorMessage({
  theme = 'error',
  message = '',
  className,
}: Props) {
  return ...
}
```

src/components/base/ErrorMessage/ErrorMessage.js
```diff
@@ -0,0 +1,34 @@
...
+      ErrorMessageStyle.core,
+      ErrorMessageStyle[theme],
+    )}
+    >
+      {message}
+    </span>
+  )
+}
+
+const MemoizedErrorMessage = React.memo/* :: <Props> */(ErrorMessage)
```
flexxnn: 👍

flow-typed/components.js
```diff
@@ -10,3 +10,8 @@ declare type AddressPickerAddress = {|
   address: Address,
   title: string,
 |}
+
+declare type StyleComponent<Theme: ?string> = {
```
stcherenkov: We changed the name to Base Component

stcherenkov: And also maybe call it `Themes` instead of `Theme`? Because there could be multiple

src/components/JTextArea/JTextAreaField.js
```diff
@@ -0,0 +1,45 @@
...
+import { JTextArea } from './JTextArea'
+
+import JTextAreaFieldStyle from './jTextAreaField.m.scss'
+
+type Props =
+  FieldRenderProps
+  & { getMeta: Function /* TODO Use mose specific typing */}
+
+export function JTextAreaField(props: Props) {
+  props.getMeta(props.meta)
```
stcherenkov: Do we really need this function?

src/components/JTextArea/JTextArea.js
```diff
@@ -0,0 +1,64 @@
...
+  if (target) {
+    while (target.clientHeight < target.scrollHeight && target.rows <= MAX_ROWS) {
+      // mutating the DOM
+      // eslint-disable-next-line fp/no-mutation, no-param-reassign
+      target.rows += 1
+    }
+  }
+}
+
+export class JTextArea extends PureComponent<Props, void> {
```
stcherenkov: Maybe we should also rename it to `JAutoExpandTextArea` or something like this, so it would not interfere with basic TextArea without auto expand function.

src/components/JTextArea/jTextArea.m.scss
```diff
@@ -0,0 +1,43 @@
...
+  border-radius: $form-border-radius;
+  line-height: $form-field-line-height;
+  resize: none;
+  border-width: 1px;
+  border-style: solid;
+  padding: $form-field-white-padding-vertical $form-field-white-horizontal-padding;
+  font-weight: 600;
+
+  // stylelint-disable selector-nested-pattern, rscss/class-format
+  &.blue {
```
IViol: can we make plain structure in css modules?
to avoid selectors like `.with.error`

src/components/WalletDataStep/WalletDataStep.js
```diff
@@ -5,6 +5,11 @@ import { t } from 'ttag'
 
 import WalletStep from 'components/WalletStep'
 import { JInput } from 'components/base'
+import { JTextArea } from 'components/JTextArea'
+import { ErrorMessage } from 'components/base/ErrorMessage'
+import { handleTargetValue } from 'utils/eventHandlers'
+
+import WalletDataStepStyle from './WallletDataStep.m.scss'
```
IViol: WalletDataStepStyle -> walletDataStepStyle

src/components/base/ErrorMessage/ErrorMessage.js
```diff
@@ -0,0 +1,34 @@
+// @flow
+
+import React from 'react'
+import classNames from 'classnames'
+
+import ErrorMessageStyle from './ErrorMessage.m.scss'
```
IViol: ErrorMessageStyle -> errorMessageStyle

src/components/base/ErrorMessage/ErrorMessage.js
```diff
@@ -0,0 +1,34 @@
...
+
+type Theme = 'info' | 'warning' | 'error'
+type Props = StyleComponent<Theme> & {
+  message: string,
+}
+
+function ErrorMessage({
```
IViol: Let's rename it :)

flow-typed/components.js
IViol: @stcherenkov https://github.com/jibrelnetwork/frontend/blob/master/codestyle.md#union-types

stcherenkov: Hm, well, good point

src/styles/mixins.scss
```diff
@@ -45,3 +45,7 @@
   padding-left: 10px;
   padding-right: 10px;
 }
+
+@mixin property-animation($property: all, $duration: 200ms) {
```
flexxnn: I think the `property-transition` is better name

src/components/JTextArea/JTextAreaField.js
```diff
@@ -1,45 +1,45 @@
...
-      {errorText && <ErrorMessage message={errorText} theme='error' />}
-    </Fragment>
+      {hasError && <JFieldMessage message={String(errorText)} theme='error' />}
+    </div>
   )
 }
 
 JTextAreaField.defaultProps = {
-  getMeta: noop,
+  offset: 'mb16',
```
IViol: probably it should be more general property, like:
```
  variants: string[],
```

src/components/JTextArea/JTextArea.js
```diff
@@ -28,37 +44,91 @@ function heightCalc({ currentTarget: target }: SyntheticEvent<HTMLTextAreaElemen
...
+        <textarea
+          {...omitedProps}
+          onChange={this.handleChange}
+          onBlur={this.handleBlur}
+          onFocus={this.handleFocus}
+          id={elementID}
+          className={classNames(jTextAreaStyle.input)}
+          value={this.props.value}
+          rows={this.props.rows}
+          disabled={this.props.disabled}
```
IViol: isn't it passed with `omitedProps`?

src/components/JTextArea/JTextArea.js
```diff
@@ -28,37 +44,91 @@ function heightCalc({ currentTarget: target }: SyntheticEvent<HTMLTextAreaElemen
...
+          </label>)}
+        <textarea
+          {...omitedProps}
+          onChange={this.handleChange}
+          onBlur={this.handleBlur}
+          onFocus={this.handleFocus}
+          id={elementID}
+          className={classNames(jTextAreaStyle.input)}
+          value={this.props.value}
+          rows={this.props.rows}
```
IViol: seems value & rows could be passed with `omitedProps`

src/components/JTextArea/JTextArea.js
```diff
@@ -28,37 +44,91 @@ function heightCalc({ currentTarget: target }: SyntheticEvent<HTMLTextAreaElemen
...
+    this.props.onBlur(event)
+  }
+
+  handleFocus = (event: SyntheticEvent<HTMLTextAreaElement>): void => {
+    this.setState({ focused: true })
+    this.props.onFocus(event)
   }
 
   render() {
     const omitedProps = omit(this.props, [
```
IViol: omited -> omitted

but would be better to name it for example restProps/otherProps/remainingProps

src/components/JTextArea/jTextArea.m.scss
```diff
@@ -16,28 +15,79 @@
...
+        color: $main-red;
+      }
+    }
+
+    &:hover,
+    &:focus {
+      box-shadow: 0 9px 18px rgba(8, 30, 57, 0.05);
     }
   }
 }
```
IViol: let's try to make selectors as plain as we can

src/components/WalletDataStep/WalletDataStep.js
```diff
@@ -42,19 +37,17 @@ const WalletDataStep = ({
...
-        value={valueData}
-        color='white'
-        name='wallet-data'
-        placeholder={t`Address, Private key, BIP32 XPUB, Mnemonic`}
-        rows={2}
-        autoFocus
-      />
-      {invalidFields.data && <ErrorMessage message={invalidFields.data} theme='error' />}
-    </Fragment>
+    <JInput
```
IViol: JInput?

src/components/JTextArea/JTextArea.js
Creepypoke: It make sense, thank you

src/components/JTextArea/jTextArea.m.scss
Creepypoke: I agree with you, but I would prefer change selectors in next iteration of this component 😅

PR #715, opened 2019-03-26
Fix functional tests runner

stcherenkov:
Update WebdriverIO to latest version
Update WebdriverIO utils and related libs
Separate config into 3 independent parts
Add local running on Chrome using ChromeDriver
Fix remote running using BrowserStack
Add example WebStorm run config
Update npm run commands
Add tests files to eslint ignore


config/wdio/base.conf.js
```diff
@@ -0,0 +1,42 @@
+const fs = require('fs-extra')
+const path = require('path')
+
+const REPORT_OUTPUT_DIR = path.resolve(__dirname, '../../reports/functional-tests')
```
Creepypoke: Maybe we can get this path from `.env` file or same, it's possible?

stcherenkov: I could not find any case for having reports directory other than predefined. Do you know such?

PR #717, opened 2019-03-28
JWWEB-694 JInputField component

flexxnn:


src/components/base/JInputField/JInputField.js
```diff
@@ -0,0 +1,116 @@
...
+          className={jInputFieldStyle.input}
+          placeholder={placeholder}
+          disabled={disabled}
+          {...input}
+          {...rest}
+        />
+        {hasTitle && (
+          <div className={jInputFieldStyle.title}>
+            {title}
+          </div>
```
IViol: better to call it label and change from div to label tag

src/utils/form/getErrorMessage.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+function getErrorMessage(meta: FinalFormMeta, validateType: ?JInputValidateType): ?string {
+  const err = meta.error || meta.submitError
+
+  const error = (typeof err === 'string')
+    ? err
+    : err instanceof Error && err.message
+      ? err.message
+      : ''
```
IViol:
```
    : err instanceof Error && err.message
      ? err.message
      : ''
```
->
```
err.message
```

src/utils/form/getErrorMessage.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+function getErrorMessage(meta: FinalFormMeta, validateType: ?JInputValidateType): ?string {
+  const err = meta.error || meta.submitError
```
IViol: maybe call it `metaError`?

it's not ok to have 2 vars `err` & `error`

metaError & strError would be great

src/utils/form/getErrorMessage.js
```diff
@@ -0,0 +1,21 @@
...
+
+  const error = (typeof err === 'string')
+    ? err
+    : err instanceof Error && err.message
+      ? err.message
+      : ''
+
+  if (validateType === 'dirtySinceLastSubmit' && !meta[validateType]) {
+    return error
+  } else if ((validateType === 'touched' || validateType === 'visited') && meta[validateType]) {
```
IViol: let's create specific vars, like isDirty, isTouched, isVisited, isExist(for meta[validateType])

currently really hard to read

src/components/base/JInputField/JInputField.js
```diff
@@ -0,0 +1,116 @@
...
+import jInputFieldStyle from './jInputField.m.scss'
+
+export type JInputType = 'text' | 'password'
+export type JInputTheme = 'white'
+export type JInputValidateType =
+  'touched' |
+  'visited' |
+  'dirtySinceLastSubmit'
+
+type Props = {
```
Creepypoke: Let's use `StyledComponent<Theme>` ?

src/components/base/JInputField/JInputField.js
```diff
@@ -0,0 +1,116 @@
...
+}
+
+function JInputField({
+  type,
+  theme,
+  validateType,
+  infoMessage,
+  title,
+  placeholder,
+  input,
```
IViol: shouldn't input prop have `defaultProps` value?

src/components/base/JInputField/JInputField.js
Creepypoke: Completely agree, also it consistent with JTextArea props

flow-typed/final-form.js
```diff
@@ -27,3 +27,7 @@ declare type FinalFormMeta = {
   valid?: boolean,
   visited?: boolean,
 }
+
+declare type FinalFormValidateType = 'touched' |
```
Creepypoke: Are you sure, that we need this type everywhere else, except JInputField component?

src/components/base/JInputField/JInputField.js
```diff
@@ -0,0 +1,111 @@
...
+  const isActive = meta.active || hasValue || (hasLabel && hasPlaceholder)
+
+  return (
+    <div
+      className={classNames(
+        jInputFieldStyle.core,
+        jInputFieldStyle[theme],
+        className,
+      )}
+      onClick={() => textInput.current.focus()}
```
Creepypoke: It's not good idea: create new function in render.
I guess, the better way is
```
onClick={textInput.current.focus}
```

src/components/base/JInputField/JInputField.js
```diff
@@ -0,0 +1,111 @@
...
+  const hasPlaceholder = !!placeholder
+  const hasError = !!errorMessage
+  const hasInfo = !!infoMessage
+  const hasValue = !!input.value
+  const isActive = meta.active || hasValue || (hasLabel && hasPlaceholder)
+
+  return (
+    <div
+      className={classNames(
+        jInputFieldStyle.core,
```
Creepypoke: Now we are adding the plain-text classes for functional tests:
Form of this classes: `__%some-element-name%`
https://github.com/jibrelnetwork/jwallet-web/pull/707/files#diff-8a1bb7f77977752c184f4c2b841e6a2fR104

src/components/base/JInputField/JInputField.js
```diff
@@ -0,0 +1,111 @@
...
+  placeholder: string,
+  type: JInputType,
+  validateType: FinalFormValidateType,
+}
+
+function JInputField({
+  className,
+  theme,
+  disabled,
+  infoMessage,
```
Creepypoke: I guess, we don't need to set some message explicitly from outside

src/components/base/JInputField/JInputField.js
```diff
@@ -0,0 +1,111 @@
...
+          {...rest}
+          {...input}
+        />
+        {hasLabel && (
+          <div className={jInputFieldStyle.label}>
+            {label}
+          </div>
+        )}
+      </div>
+      {!hasError && hasInfo && (
```
Creepypoke: Can we have double message in same time?
If don't, then I would be want divide view logic with business logic via set props above.

flow-typed/final-form.js
flexxnn: I think we need this functionality in all fields

src/components/base/JInputField/JInputField.js
flexxnn: [eslint] Handler function for onClick prop key must begin with 'handle' (react/jsx-handler-names)
This will fail with this eslint error

src/components/base/JInputField/JInputField.js
flexxnn: It is input info message. We have this in design and we need it

src/components/base/JInputField/JInputField.js
stcherenkov: This is an error. You should definitely not pass new function to HTML element on every render. You could do it in several ways so there should be no ESLint error, or you could turn ESLint rule off, but in general we prefer writing error-prone code to writing code exactly according to rules.

flexxnn: Fixed in this commit: https://github.com/jibrelnetwork/jwallet-web/pull/717/commits/1a0a767028c60ac2c2def854a76795335010b1b6

PR #718, opened 2019-03-28
Add fetching status of password indicator

IViol:


src/components/PasswordField/Indicator/indicator.m.scss
```diff
@@ -0,0 +1,91 @@
+@import "styles/colors";
+@import "styles/mixins";
+
+$password-field-indicator-height: 6px;
+
+.core {
+  all: initial;
```
Creepypoke: I think, we no longer use `all: initial` in styles.

PR #719, opened 2019-03-29
JWWEB-715: Update MenuLayout

stcherenkov:
MenuLayout now takes only view name and renders menu bar using internal config

Also:
Migrate to new routes list
Add minimal responsive css for MenuLayout


src/layouts/MenuLayout/components/MainMenu.js
```diff
@@ -0,0 +1,49 @@
...
+}: Props) {
+  return (
+    <header
+      className={classnames(
+        '__mainmenu',
+        menuLayoutStyle.menu,
+        menuLayoutStyle[mode],
+      )}
+    >
+      {previousRouteNameFallback && (
```
Creepypoke: Maybe add markup like `nav > ol > li` for navigation items?

src/layouts/MenuLayout/menuLayout.m.scss
```diff
@@ -0,0 +1,35 @@
+@import 'styles/mixins';
+@import 'styles/colors';
```
Creepypoke: Colors already include in mixins.

PR #724, opened 2019-04-01
JWWEB-733: Add transfer buttons to home screen

stcherenkov:
Rename DigitalAssetsGridView and DigitalAssetsGridViewContainer components to HomeTemplate and Home respectively
Add required icons


src/pages/Home/HomeTemplate.js
```diff
@@ -51,18 +54,61 @@ class DigitalAssetsGridView extends Component<Props> {
...
+                  color='blue'
+                />
+              </div>
+              {t`Receive`}
+            </JLink>
+            <JLink
+              className={style.transferButton}
+              href={JCASH_UTM_URL}
+            >
+              <div className={style.transferButtonIcon}>
```
IViol: `div` inside `a` is not semantic

src/pages/Home/HomeTemplate.js
```diff
@@ -13,9 +13,12 @@ import {
   DigitalAssetsFilter,
 } from 'components'
 
+import style from './home.m.scss'
```
IViol: style -> homeStyle

src/pages/Home/HomeTemplate.js
```diff
@@ -51,18 +54,61 @@ class DigitalAssetsGridView extends Component<Props> {
     const filterCount: number = filterOptions.isHideZeroBalance ? 1 : 0
 
     return (
-      <div className='digital-assets-grid-view'>
-        <div className='header'>
-          <div className='container'>
+      <div className={style.home}>
```
IViol: style.home -> homeStyle.core

src/pages/Home/HomeTemplate.js
```diff
@@ -96,4 +142,4 @@ class DigitalAssetsGridView extends Component<Props> {
   }
 }
 
-export default DigitalAssetsGridView
+export default HomeTemplate
```
IViol: export default?

src/pages/Home/home.m.scss
```diff
@@ -0,0 +1,106 @@
...
+
+.title {
+  padding: 54px 0 24px;
+  font-size: 24px;
+  line-height: 34px;
+  color: $main-black;
+  font-weight: normal;
+}
+
+.transfer-buttons {
```
IViol: why not just `buttons` or `list` or `actions`, whatever in one word?

src/pages/Home/home.m.scss
```diff
@@ -0,0 +1,106 @@
...
+  @include flex-row;
+
+  margin: 0 -16px;
+}
+
+.transfer-button {
```
IViol: why need to repeat "transfer" in several selectors?

src/pages/Home/home.m.scss
stcherenkov: Because there are multiple attention locuses on the page, and I need to specify which area is it about. I don't know now whether we would have other buttons on page encapsulated into separate components or not (it's depends of specific implementation a lot), so I made transfer buttons separated by default.

I will rename them, but I think it's ok to have namespaced elements if they are in common context.

Btw, WebStorm now supports converting `kebab-case-class-names` to `camelCaseClassNames` for CSS modules in suggestions with zero configuration. Try it!))

src/pages/Home/HomeTemplate.js
stcherenkov: It is considered ok since HTML 5 spec: https://www.w3.org/TR/2011/WD-html5-20110525/text-level-semantics.html#the-a-element

src/pages/Home/home.m.scss
```diff
@@ -0,0 +1,106 @@
+@import 'styles/mixins';
+@import 'styles/colors';
```
Creepypoke: colors already included in mixins

PR #727, opened 2019-04-02
Update MenuPanel

IViol:


src/components/MenuPanel/Actions.js
```diff
@@ -0,0 +1,68 @@
...
+  JIcon,
+  JLink,
+} from 'components/base'
+
+import menuPanelStyle from './menuPanel.m.scss'
+
+export function Actions() {
+  return (
+    <nav className={menuPanelStyle.actions}>
+      <ol>
```
stcherenkov: `ul` please

There is no order here, all navigation items are equal

stcherenkov:
All the icons require `-use-fill`-postfix

src/components/MenuPanel/Actions.js
```diff
@@ -0,0 +1,68 @@
...
+        <li>
+          <JLink
+            href='/'
+            className={menuPanelStyle.action}
+            activeClassName={menuPanelStyle.active}
+          >
+            <JIcon name='home' size='medium' />
```
stcherenkov: Also no size please. Icon already contains its size, there is no need to specify it directly.

src/components/MenuPanel/MenuPanel.js
```diff
@@ -1,119 +1,66 @@
...
-      isActionsMoreActive: !this.state.isActionsMoreActive,
-    })
-  }
-
-  toggleWalletManager = () => {
-    this.setState({
-      isActionsMoreActive: false,
-      isWalletManagerActive: !this.state.isWalletManagerActive,
-    })
+export class MenuPanel extends PureComponent<Props> {
```
stcherenkov: I suggest we put `MenuPanel` under `MenuLayout`, because it is the one and only place we would use it.

src/components/MenuPanel/MenuPanel.js
```diff
@@ -1,119 +1,66 @@
...
+        <Top
+          walletName={walletName}
+          fiatCurrency={fiatCurrency}
+          mnemonicAddressName={mnemonicAddressName}
+          fiatBalance={fiatBalance}
+          isMnemonic={isMnemonic}
+        />
+        <Separator />
+        <Actions />
+        <Separator />
```
stcherenkov: Why do you put separator as separate component? It is purely decorative entity not really useful by itself, only as part of something (like `Top`, or `Actions`, or `Settings`) and can be even done without using any HTML (`:after` of `border`)

src/components/MenuPanel/MenuPanel.js
```diff
@@ -53,7 +59,7 @@ export class MenuPanel extends PureComponent<Props> {
           isMnemonic={isMnemonic}
         />
         <Separator />
-        <Actions />
+        <Actions routeName={routeName} />
```
stcherenkov: The same hack is required for Settings. :( Are you sure you want to put them into separate components?

IViol: no need to do it for settings

src/components/DigitalAssetsGrid/AddAsset/AddAsset.js
```diff
@@ -0,0 +1,24 @@
+// @flow
```
Creepypoke: Why renamed file looks like new file in git?

src/components/base/JLink/JLinkInternal.js
```diff
@@ -13,45 +14,31 @@ export type JLinkInternalProps = JLinkProps & {
...
-      {...props}
+      {...omit(props, ['href'])}
       routeName={route.name}
       routeParams={route.params}
     />
   )
 }
 
-export const JLinkInternal = withRouter(JLinkInternalDisconnected)
+const JLinkInternalEnhanced = withRouter(JLinkInternal)
```
Creepypoke: Maybe use `Reac.memo` for "dump" components?

src/layouts/MenuLayout/components/MenuPanel/components/Action.js
```diff
@@ -0,0 +1,37 @@
...
+  +iconName: string,
+|}
+
+export function Action({
+  href,
+  label,
+  iconName,
+}: Props) {
+  return (
+    <li>
```
Creepypoke: If we will use this component only in context `Actions` component, it's fine.
Otherwise, will better, if we don't make guess-work about context where component will be put.

src/components/DigitalAssetsGrid/AddAsset/AddAsset.js
IViol: it was not only renamed

src/routes/Transactions/routes/Asset/TransactionsAssetView.js
```diff
@@ -123,7 +123,7 @@ function TransactionsAssetView({
             </div>
             <div className='send' title={t`Send asset`}>
               <JFlatButton
-                to={`/digital-assets/send?asset=${asset.blockchainParams.address}`}
+                to={`/send?asset=${asset.blockchainParams.address}`}
```
flexxnn: Do we have check before, that `asset.blockchainParams.address` exists?

src/utils/wallets/getMnemonicAddressName.js
```diff
@@ -0,0 +1,28 @@
...
+  const {
+    type,
+    addressIndex,
+  }: Wallet = wallet
+
+  const isMnemonic: boolean = checkMnemonicType(type)
+  const index: number = addressIndex + 1
+
+  if (!isMnemonic) {
+    throw new Error(t`WalletInconsistentDataErrorData`)
```
flexxnn: Now, you can change this into `WalletInconsistentDataError`

`JWWEB-621: Selectors, that throw`, merged :)

src/components/base/JIcon/JIcon.js
```diff
@@ -1,45 +1,51 @@
...
-        height={iconData.height}
-      >
-        <use xlinkHref={iconData.url} />
-      </svg>
-    )
-  } else {
-    return (
-      <div className='j-icon -no-icon' />
-    )
+class JIcon extends PureComponent<Props> {
```
flexxnn: https://www.peterbe.com/plog/memo-instead-of-purecomponent

I think it's better to use React.memo instead of a PureComponent  here

src/components/DigitalAssetsGrid/AddAsset/AddAsset.js
```diff
@@ -0,0 +1,24 @@
...
+  JLink,
+} from 'components/base'
+
+export function AddAsset() {
+  return (
+    <JLink
+      href='/assets'
+      className='add-asset'
+    >
+      <div className='icon'>
```
flexxnn: JLink is `<a` that semantic inline element, so I think it is better to use `span` here

src/layouts/MenuLayout/components/MenuPanel/menuPanel.m.scss
```diff
@@ -0,0 +1,164 @@
...
+}
+
+.core,
+.top,
+.wrapper,
+.actions,
+.action,
+.action svg,
+.back {
+  @include hover-animation;
```
flexxnn: Have a look to rename this to `hover-transition` (in future)

src/layouts/MenuLayout/components/MenuPanel/tests/MenuPanel.test.js
```diff
@@ -0,0 +1,298 @@
...
+
+import React from 'react'
+import { shallow } from 'enzyme'
+
+import {
+  divDecimals,
+  formatBalance,
+} from 'utils/numbers'
+
+jest.mock('../../../../../utils/sprite/spriteUI', () => ({
```
flexxnn: It is possible to mock this before all unit tests
https://github.com/jibrelnetwork/jwallet-web/blob/380e03a7537484f9d7dc4a4898b24c21f25fa7c8/.jest/register-context.js
https://github.com/jibrelnetwork/jwallet-web/blob/380e03a7537484f9d7dc4a4898b24c21f25fa7c8/jest.config.js#L36

It's done in my PR

src/layouts/MenuLayout/components/MenuPanel/menuPanel.m.scss
```diff
@@ -0,0 +1,164 @@
+@import 'styles/colors';
+@import 'styles/mixins';
```
Creepypoke: `mixins` already include `colors` in depth

src/layouts/MenuLayout/menuLayout.m.scss
```diff
@@ -1,35 +1,26 @@
+@import 'styles/base';
 @import 'styles/mixins';
```
Creepypoke: same, `mixins` already include base

src/components/DigitalAssetsGrid/AddAsset/AddAsset.js
Creepypoke: +1
`span` is good choice, but, meanwhile, in HTML5 specs content type for anchors is don't matter.
https://www.w3.org/TR/2011/WD-html-markup-20110405/a.html#a-content-model

src/routes/Transactions/routes/Asset/TransactionsAssetView.js
Creepypoke: I guess, we need to think about more stable (without "maybe" fields) and highly specialized record structures, in future

stcherenkov: Users can change the link and set the `asset` value to `blahblah`, so the check should be definitely placed on target page, not here.

PR #728, opened 2019-04-03
add option for gzip_static,gzip_buffers,gunzip,gunzip_buffers

VolverinPro:


stcherenkov:
Better to target `develop`, I guess?

VolverinPro:
@stcherenkov , yes. You are right.

PR #729, opened 2019-04-03
Feature/enable static gzip

VolverinPro:
Add some options of nginx for gzip_static , gzip_buffers
http://nginx.org/ru/docs/http/ngx_http_gzip_static_module.html
http://nginx.org/ru/docs/http/ngx_http_gzip_module.html#gzip_buffers


VolverinPro:
According to github flow, when creating a feature, the master branch is taken as the basis, but because we create PR on develop , it is therefore more correct to base the develop branch.

PR #730, opened 2019-04-04
add option for gzip_static,gz_buffers [DEVOPS-251]

VolverinPro:
Add some options of nginx for gzip_static , gzip_buffers
http://nginx.org/ru/docs/http/ngx_http_gzip_static_module.html
http://nginx.org/ru/docs/http/ngx_http_gzip_module.html#gzip_buffers


nginx.conf
```diff
@@ -27,11 +27,13 @@ http {
     server_tokens off;
 
     gzip on;
+    gzip_buffers 64 8k; # http://nginx.org/ru/docs/http/ngx_http_gzip_module.html#gzip_buffers
```
soar: Why do you use 64/8k buffers?

nginx.conf
```diff
@@ -27,11 +27,13 @@ http {
...
     gzip_comp_level 6;
     gzip_disable "msie6";
     gzip_min_length 1000;
     gzip_proxied any;
     gzip_types *;
+    gzip_static  on;  # http://nginx.org/ru/docs/http/ngx_http_gzip_static_module.html
```
soar: Remove the extra space, please

soar:
As I see, there are only gzipped versions for `js` and `css` files. For example, `svg`'s in `/app/assets/icons/large/` are not compressed. What about them? @stcherenkov

stcherenkov:
We have several recommended rules for compression [configured here](https://github.com/jibrelnetwork/jwallet-web/blob/develop/webpack.config.js#L444-L450) with [defaults described here](https://github.com/webpack-contrib/compression-webpack-plugin):
- .map-files (mappings from non-minified source files to minified production files) are not processed, because browser does not load them, until put into development mode
- files less than 8kb are not processed, because http package is 8kb, so server will send them entirely in one package anyway
- files that are more than 80% of initial size after compression are not processed, because we are slowing down browser more than actually save on loading speed

And also we use zopfli algorithm with compression level 9 (that's kinda cool, cause you can't do such good compression fast enough to be on-the-fly with nginx)

stcherenkov:
Also for svgs we migrate from separate files to svg sprite that builds with our app, so we don't need to gzip those.

stcherenkov:
And about PR: does our nginx container already contain required plugin? Don't we need to include it somehow?

nginx.conf
VolverinPro: According to the description in nginx, I assumed that the greater the number of gzip buffers, the faster nginx will be able to respond to the client. It's better to have more small buffers available than a few huge buffers.

soar:
If we talk about the compression level, it's tricky. For example, AWS CloudFront uses the 6th level for their static content, because it has the best compression quality to decompression speed ratio. The 9th has a smaller size, but the client will have a longer decompression stage. @stcherenkov

soar:
> files less than 8kb are not processed, because http package is 8kb, so server will send them entirely in one package anyway

Please, give me more information about this. I am not sure that you are right.

stcherenkov:
> > files less than 8kb are not processed, because http package is 8kb, so server will send them entirely in one package anyway
> 
> Please, give me more information about this. I am not sure that you are right.

Well, looks like I'm really not: [wiki](https://en.wikipedia.org/wiki/Maximum_transmission_unit) says it depends on network type. But it is safe to assume that files less than 2kb require at least one http packet so we don't need to compress them.

But if we are going to have http/2, we should compress everything. Do we?

stcherenkov:
> If we talk about the compression level, it's tricky. For example, AWS CloudFront uses the 6th level for their static content, because it has the best compression quality to decompression speed ratio. The 9th has a smaller size, but the client will have a longer decompression stage. @stcherenkov

Do they have maybe a blog post about it? Everything I found says decompression rate is mostly the same with losses around 1-1.5% (for the load speed we can ignore it as it is less than measurement error).

nginx.conf
```diff
@@ -46,7 +48,13 @@ http {
             add_header Content-Security-Policy $CSP;
             root /app;
             index index.html;
-            try_files $uri /index.html;
+            try_files $uri $uri/ /index.html;
+        }
+
+        location /storybook {
```
stcherenkov: Uhm, thanks, but I don't want to put storybook to production server))

soar:
Yes, our servers and our CDN support HTTP/2, so we shouldn't think about separate files size. I've found some investigations from different companies, that it's pointless to compress files less than 850 bytes - they become larger after compression. So, maybe it will be a good idea to compress only files larger than 1kb. @stcherenkov

nginx.conf
soar: That's the wrong push from a different task, it will be removed here, sorry for the mistake

stcherenkov:
> Yes, our servers and our CDN support HTTP/2, so we shouldn't think about separate files size. I've found some investigations from different companies, that it's pointless to compress files less than 850 bytes - they become larger after compression. So, maybe it will be a good idea to compress only files larger than 1kb. @stcherenkov

Ok, I'll fix it in this PR

VolverinPro:
> And about PR: does our nginx container already contain required plugin? Don't we need to include it somehow?

Yes, we do a check of nginx.conf when building docker image.
I checked work of plugin nginx_static, examples:
a) curl --silent --write-out 'size_download=%{size_download}\n'  --http1.0  -H "Accept-Encoding: gzip,deflate"  --output /dev/null  http://127.0.0.1:8080/bc36b4999f8e54110c8c.worker.js
size_download=978801     - when client doesn't support compression.
b) curl --silent --write-out 'size_download=%{size_download}\n' -H "Accept-Encoding: gzip,deflate"  --output /dev/null  http://127.0.0.1:8080/bc36b4999f8e54110c8c.worker.js
size_download=280762   - - when client support compression.

stcherenkov:
> But if we using below script, we have get almost the same time for both responses.

I guess it's curl internal cache?

stcherenkov:
Fixed webpack config

soar:
изображение

Interesting thing: this file has length of 286 bytes but also 621 bytes of headers are passed with it from server to client.

VolverinPro:
> > But if we using below script, we have get almost the same time for both responses.
> 
> I guess it's curl internal cache?
I checked with curl option no-cache and results are not the same %)
a) curl -s -w 'size_download=%{size_download}\nTotal: %{time_total}s\n' --http1.0 -H 'Accept-Encoding: gzip,deflate' -H 'Cache-Control: no-cache' --output /dev/null http://127.0.0.1:8080/bc36b4999f8e54110c8c.worker.js
size_download=978801
Total: 0.004014s
when client doesn't support compression.
b) curl -s -w 'size_download=%{size_download}\nTotal: %{time_total}s\n' -H 'Accept-Encoding: gzip,deflate' -H 'Cache-Control: no-cache' --output /dev/null http://127.0.0.1:8080/bc36b4999f8e54110c8c.worker.js
size_download=280762
Total: 0.001266s
when client support compression.

But if we using below script,we have get almost the same time fot both responce.
#!/bin/bash
for i in {1..100}
do
START=$(date +%s)
curl -s -w 'size_download=%{size_download}\nTotal: %{time_total}s\n' --http1.0 -H 'Accept-Encoding: gzip,deflate' -H 'Cache-Control: no-cache' --output /dev/null "http://127.0.0.1:8080/bc36b4999f8e54110c8c.worker.js"
curl -s -w 'size_download=%{size_download}\nTotal: %{time_total}s\n' -H 'Accept-Encoding: gzip,deflate' -H 'Cache-Control: no-cache' --output /dev/null "http://127.0.0.1:8080/bc36b4999f8e54110c8c.worker.js"
END=$(date +%s)
DIFF=$(( $END - $START ))
echo "It took $DIFF seconds"
done

soar:
@VolverinPro Why they should be different?

VolverinPro:
> > But if we using below script, we have get almost the same time for both responses.
> 
> I guess it's curl internal cache?

I checked with curl option no-cache and results are not the same %)
#!/bin/bash
for i in {1..100}
do
START=$(date +%s)
curl -s -w 'size_download=%{size_download}\nTotal: %{time_total}s\n' --http1.0 -H 'Accept-Encoding: gzip,deflate' -H 'Cache-Control: no-cache' --output /dev/null "http://127.0.0.1:8080/bc36b4999f8e54110c8c.worker.js"
curl -s -w 'size_download=%{size_download}\nTotal: %{time_total}s\n' -H 'Accept-Encoding: gzip,deflate' -H 'Cache-Control: no-cache' --output /dev/null "http://127.0.0.1:8080/bc36b4999f8e54110c8c.worker.js"
END=$(date +%s)
DIFF=$(( $END - $START ))
echo "It took $DIFF seconds"
done

VolverinPro:
> @VolverinPro Why they should be different?
I supose, because gz files should sended of  nginx more faster than simple files

soar:
> > @VolverinPro Why they should be different?
> > I supose, because gz files should sended of  nginx to client  more faster than simple files

I don't know what you are expecting to see, but I think difference in two times is quite enough:

```
#!/bin/sh
for i in {1..100}
do
START=$(date +%s)
curl -s -w 'size_download=%{size_download}\nTotal: %{time_total}s\n' -H 'Cache-Control: no-cache' --output /dev/null "http://127.0.0.1:80/1.74d1f103295f141f7a4a.worker.js"
curl -s -w 'size_download=%{size_download}\nTotal: %{time_total}s\n' -H 'Accept-Encoding: gzip,deflate' -H 'Cache-Control: no-cache' --output /dev/null "http://127.0.0.1:80/1.74d1f103295f141f7a4a.worker.js"
END=$(date +%s)
DIFF=$(( $END - $START ))
echo "It took $DIFF seconds"
done
```

```
size_download=818677
Total: 0.004082s
size_download=379707
Total: 0.002518s
```

VolverinPro:
> > > @VolverinPro Why they should be different?
> > > I supose, because gz files should sended of  nginx to client  more faster than simple files
> 
> I don't know what you are expecting to see, but I think difference in two times is quite enough:
> 
> ```
> #!/bin/sh
> for i in {1..100}
> do
> START=$(date +%s)
> curl -s -w 'size_download=%{size_download}\nTotal: %{time_total}s\n' -H 'Cache-Control: no-cache' --output /dev/null "http://127.0.0.1:80/1.74d1f103295f141f7a4a.worker.js"
> curl -s -w 'size_download=%{size_download}\nTotal: %{time_total}s\n' -H 'Accept-Encoding: gzip,deflate' -H 'Cache-Control: no-cache' --output /dev/null "http://127.0.0.1:80/1.74d1f103295f141f7a4a.worker.js"
> END=$(date +%s)
> DIFF=$(( $END - $START ))
> echo "It took $DIFF seconds"
> done
> ```
> ```
> size_download=818677
> Total: 0.004082s
> size_download=379707
> Total: 0.002518s
> ```

I agree %)

PR #731, opened 2019-04-04
Storybook Build [DEVOPS-167]

skydivervrn:


run.sh
```diff
@@ -2,6 +2,12 @@
 
 RUNMODE="${1:-start}"
 
+if [ "$ENV" = "demo" ]; then
+    sed -i '/location \/status/i \        location /storybook {\n            root /docs; \n            try_files $uri $uri/ /index.html; \n        }\n' /etc/nginx/nginx.conf
```
soar: Can we make this line more readable? Maybe create subconfig with this line which will be included here?

run.sh
```diff
@@ -2,6 +2,12 @@
...
+    echo $ENV
```
soar: It's better to make this like look like:
```
echo "Starting server in ${ENV:-normal} mode"
```

soar:
Much better and it works 👍 . But config is little bit ugly in patched line. I understand that it's just "temporary" file, but it would be great if you can make it prettier and more aligned.

изображение

PR #737, opened 2019-04-10
JWWEB-747: Unexpected Error handling

stcherenkov:
Looks like nobody generated .po file for a long time...


src/AppRouter.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import React from 'react'
+import React, { Component } from 'react'
```
Creepypoke: I guess, we can use `PureComponent` here

PR #739, opened 2019-04-11
JWWEB-729 wallets start screen

IViol:


src/components/LogoHeader/tests/LogoHeader.test.js
```diff
@@ -0,0 +1,42 @@
...
+            matchUrl: () => null,
+          },
+        },
+      )
+    }
+  },
+}))
+
+// eslint-disable-next-line import/first
+import { LogoHeader } from '../LogoHeader'
```
Creepypoke: Maybe better will move mocks under imports, in future?

src/components/NewWalletButtons/tests/NewWalletButtons.test.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+})
+
+// eslint-disable-next-line import/first
+import { NewWalletButtons } from '../NewWalletButtons'
+
+const MOCK_EVENT = { preventDefault: () => {} }
+
+describe('NewWalletButtons', () => {
```
stcherenkov: Very strange test.

We are testing visual component, but it requires web worker to run?

We check texts? What if they change? Test will fail? Why? Implementation is still absolutely correct.

src/pages/WalletsStart/WalletsStart.js
```diff
@@ -0,0 +1,69 @@
...
+  LogoHeader,
+  NewWalletButtons,
+} from 'components'
+
+import walletsStartStyle from './walletsStart.m.scss'
+
+export type WalletAction = 'CREATE' | 'IMPORT'
+type WalletActions = { [WalletAction]: WalletAction }
+
+type Props = {||}
```
Creepypoke: I'm not sure, but I think that `null` is better, because `null` comparing is simpler.

src/components/base/JLogo/tests/JLogo.test.js
```diff
@@ -0,0 +1,68 @@
...
+    const jLink = wrapper.childAt(0)
+
+    expect(jLink.prop('href')).toBe('/')
+    expect(jLink.hasClass('__j-logo')).toBe(true)
+  })
+
+  test('renders (specified theme)', () => {
+    const wrapper = mount(<JLogo theme='blue' />)
+
+    expect(wrapper.prop('theme')).toBe('blue')
```
stcherenkov: What is tested here? If prop is specified, then it is specified. This is tested by React core members, no need to reproduce it.

src/components/NewWalletButtons/NewWalletButtons.js
```diff
@@ -1,40 +1,52 @@
...
+  handleClick = (action: WalletAction) => (event: SyntheticEvent<HTMLDivElement>) => {
+    event.preventDefault()
+
+    this.props.onClick(action)
+  }
+
+  render() {
+    return (
+      <div className={`__new-wallet-buttons ${newWalletButtonsStyle.core}`}>
+        <JLink
```
Creepypoke: Not in this pull request, but I clearly sure that we need to add into `JRaisedButton` `JLink` modification by `href/to` props and add the new theme.
@edanilov @flexxnn @IViol What do you think about this?

PR #741, opened 2019-04-12
JWWEB-629: Add browserslist config

stcherenkov:


config/wdio/browserstack.conf.js
```diff
@@ -2,6 +2,8 @@ const fs = require('fs')
 const path = require('path')
 const ip = require('ip')
 
+const { getBrowserStackCapabilitiesByConfig } = require('@jibrelnetwork/browserslist-config')
```
Creepypoke: wow, browserlist is sharing in public?

stcherenkov: Yeah, that info is no secret: if user's browser does not support required features, we will show them recommended browsers list, based on our browserslist config

PR #743, opened 2019-04-15
Upload and download IDB snapshots for tests

stcherenkov:
To download IDB snapshot you just open your browser console and copy-paste-enter contents of `tests/utils/downloadDBSnapshot.client.js`.

To upload IDB snapshot in test, you do:
```
const uploadDBSnapshot = require('./utils/uploadDBSnapshot')

describe('Initial app loading', () => {
  it('starts with predefined account', () => {
    browser
      .url('/')
    
    uploadDBSnapshot(require('./path/to/snapshot.json'))

    // you can also make all agreements checked
    require('./utils/confirmAllAgreements')()

    // you need to reload page after all that
    browser.url('/')

    // now you are all set up, go!
    ...
  })
})
```


Creepypoke:
I think, we need to add some information (from the pr description for example) about tests into documentation.

PR #746, opened 2019-04-17
Update JAssetSymbol

Creepypoke:
View for digital assets without icon was updated according to design


src/components/base/JAssetSymbol/jAssetSymbol.scss
```diff
@@ -15,16 +15,23 @@
...
-    display: block;
-    content: attr(data-symbol);
     position: absolute;
-    top: 50%;
-    left: 50%;
-    transform: translate(-50%, -50%);
+    box-sizing: border-box;
+    content: attr(data-symbol);
+    overflow: hidden;
+    width: 40px;
```
stcherenkov: Is it always 40px? What about Asset Picker?

src/components/base/JAssetSymbol/jAssetSymbol.scss
```diff
@@ -15,16 +15,23 @@
...
+    height: 40px;
+    border: 1px solid $main-black-10;
+    border-radius: 50%;
+    font-style: normal;
     font-weight: 600;
-    font-size: inherit;
+    text-align: center;
+    text-transform: uppercase;
+    text-overflow: ellipsis;
```
stcherenkov: Didn't we decide earlier there is no ellipsis? Maybe better to hide overflow symbols?

src/components/base/JAssetSymbol/jAssetSymbol.scss
Creepypoke: It depends on version of Asset Picker do you ask :D

src/components/base/JAssetSymbol/jAssetSymbol.scss
Creepypoke: It just a defense by dumb situations with too long texts. But, I guess, your proposal could better

src/components/base/JAssetSymbol/JAssetSymbol.js
```diff
@@ -11,26 +11,35 @@ type JAssetSymbolColor = 'blue' | 'gray' | 'white'
...
+  symbol, color, isCustom,
 }: Props) {
   const url = iconsAsset[`${symbol.toLowerCase()}-usage`]
 
   return (
     <div
-      className={classNames(`j-asset-symbol ${getSymbolClassName(symbol)} -${color}`)}
-      data-symbol={symbol}
+      className={classNames(`j-asset-symbol ${getSymbolClassName(symbol, isCustom)} -${color}`)}
+      data-symbol={symbol[0]}
```
stcherenkov: Final business resolution:
- if there is less than 5 chars, display all of them
- if there are 5 chars or more, display first 3 only

Creepypoke: 🔢

PR #747, opened 2019-04-17
JWWEB-736 Stories for transaction item

Creepypoke:
- New story for single transaction was added
- Tiny style fixes
- Titles for long text was added to title and note area


src/components/TransactionItemNew/TransactionsItem.stories.js
```diff
@@ -85,8 +87,45 @@ class TransactionsList extends PureComponent<ListProps, ListState> {
   }
 }
 
+const defaultTransaction = getTransaction()
+const STATUSES = {
```
stcherenkov: Btw, those can be set as array of strings, I checked.

PR #750, opened 2019-04-19
Remove min width for tablets support

stcherenkov:


stcherenkov:
Whoops, also assets submodule changed... Sorry, forgot to remove it.

PR #761, opened 2019-04-23
Update dependencies to partially remove warnings

stcherenkov:


flexxnn:
I think we can update flow to 0.97 too
https://github.com/facebook/flow/releases/tag/v0.97.0

stcherenkov:
> I think we can update flow to 0.97 too
> https://github.com/facebook/flow/releases/tag/v0.97.0

Done

PR #768, opened 2019-04-24
Implement introduction screen

Dragonesis:


src/styles/components/jText.m.scss
```diff
@@ -7,6 +7,10 @@
   transition: color 0.3s ease 0s, opacity 0.3s ease 0s, font-size 0.3s ease 0s;
 }
 
+.main-black {
```
stcherenkov: Why is JText still alive?

Dragonesis: I missed the moment when he was completely abandoned. Even from using its classes

src/pages/Introduction/IntroductionView.js
```diff
@@ -0,0 +1,149 @@
...
+
+import introductionViewStyle from './introductionView.m.scss'
+import './swiper.scss'
+
+const sliders = [
+  {
+    id: 0,
+    img: '/assets/feature/jwallet.png',
+    animationName: 'jwallet',
+    title: 'Jwallet',
```
flexxnn: Add ttag here

src/pages/Introduction/components/Slide.js
```diff
@@ -0,0 +1,102 @@
...
+import classNames from 'classnames'
+
+import { t } from 'ttag'
+
+import jTextStyle from 'styles/components/jText.m.scss'
+
+import introductionStyle from './slide.m.scss'
+
+type Props = {|
+  id: number,
```
flexxnn: Add + here

Creepypoke:
Update submodules by `git submodule update` and `git submodule sync`, please

src/pages/Introduction/IntroductionView.js
```diff
@@ -0,0 +1,175 @@
...
+
+    return (
+      <div className={`__introduction ${introductionViewStyle.core}`} >
+        <ViewSlider
+          renderView={renderView}
+          numViews={5}
+          activeView={activeId}
+        />
+        <div className={introductionViewStyle.paginations}>
+          <button
```
stcherenkov: These buttons are created not based on data, like slides, but they represent the same data. This can lead to errors (added new feature to data list, but did not add button).

src/pages/Introduction/IntroductionView.js
```diff
@@ -0,0 +1,175 @@
...
+            type='button'
+            className={classNames(
+              introductionViewStyle.item,
+              activeId === 0 && introductionViewStyle['-current'],
+            )}
+            onClick={this.setActiveId(0)}
```
stcherenkov: Bad practice that leads to re-render every time.

You should create simple component that gets `id` and `onClick` as props, then does smth like this inside:

```
const SliderButton extends PureComponent {
  onClick = () => {
    this.props.onClick(this.props.id)
  }

  render() {
    return <button onClick={this.onClick} />
  }
}
```

This does not make any memory leaks for sure.

src/pages/Introduction/IntroductionView.js
```diff
@@ -0,0 +1,175 @@
...
+  activeId: number,
+|}
+
+const sliders = [
+  {
+    img: '/assets/feature/jwallet.png',
+    animationName: 'jwallet',
+    title: t`Jwallet`,
+    descr: [
+      t`A simple, fast and secure mobile wallet`,
```
stcherenkov: For proper translation it is better to have `<br>` inside and set the value with `dangerouslySetInnerHTML`

src/pages/Introduction/IntroductionView.js
```diff
@@ -0,0 +1,175 @@
...
+    this.setState({ activeId: number })
+  }
+
+  render() {
+    const {
+      activeId,
+    }: State = this.state
+
+    return (
+      <div className={`__introduction ${introductionViewStyle.core}`} >
```
stcherenkov: There is a ready for use responsive layout btw. See: https://github.com/jibrelnetwork/jwallet-web/tree/develop/src/layouts/StartLayout

src/pages/index.js
```diff
@@ -36,8 +36,9 @@ export { NotFound } from './NotFound/NotFound'
 export { SetPassword } from './SetPassword/SetPassword'
 export { WalletsStart } from './WalletsStart/WalletsStart'
 export { AgreementsView } from './Agreements/AgreementsView'
-
+export { IntroductionView } from './Introduction/IntroductionView'
```
IViol: should be named just `Introduction`

src/pages/Introduction/components/SliderButton.js
```diff
@@ -0,0 +1,35 @@
...
+
+type Props = {|
+  +slideId: number,
+  +isActive: boolean,
+  +onClick: (slideId: number) => any,
+|}
+
+export class SliderButton extends PureComponent<Props> {
+  handleClick = () => {
+    this.props.onClick(this.props.slideId)
```
IViol: this action should be ignored, if the target is currently active

src/pages/Introduction/components/Slide.js
```diff
@@ -0,0 +1,79 @@
...
+
+    this.state = {
+      animationData: null,
+    }
+
+    Promise
+      .all([
+        import('lottie-web'),
+        import(`../animations/${this.props.animationName}`),
+      ])
```
IViol: wouldn't it better to split loading of lib and json?

src/pages/index.js
Dragonesis: Why?

stcherenkov: Naming convention, see here: https://jibrelnetwork.atlassian.net/wiki/spaces/~40436670/pages/803766412/Routes

Dragonesis: Rename all files and lines as "Introduction"?

stcherenkov: We agreed on naming as follows:
- if the page requires separation to connector and view, we name view `SomethingView` and connector (cause it is actual page) – `Something`
- if the page does not require separation, we name the only file as `Something`

Everything correct, @IViol ?

src/pages/Introduction/Introduction.js
```diff
@@ -0,0 +1,130 @@
...
+import IntroductionStyle from './introduction.m.scss'
+
+export type Props = {|
+|}
+
+type ComponentState = {|
+  +activeId: number,
+|}
+
+const sliders = [
```
stcherenkov: `slides`?

src/pages/Introduction/Introduction.js
```diff
@@ -0,0 +1,130 @@
...
+  {
+    img: '/assets/feature/manage.png',
+    animationName: 'manage',
+    title: t`Manage`,
+    descr: t`Enjoy complete control over your digital assets. 
+      <br> Manage ETH and all ERC-20 tokens.`,
+  },
+]
+
+const renderView = ({
```
stcherenkov: So, you can actually put slides data into separate file, e.g. `data.js`, and move this function inside `ViewSlider` component, eliminating the function purpose, yes?

src/pages/Introduction/Introduction.js
```diff
@@ -0,0 +1,130 @@
...
+    const {
+      activeId,
+    } = this.state
+
+    return (
+      <StartLayout className='__introduction'>
+        <ViewSlider
+          renderView={renderView}
+          numViews={sliders.length}
+          activeView={activeId}
```
stcherenkov: If it is `id`, please call it `id`. Intermediary renaming is redundant.

src/pages/Introduction/Introduction.js
```diff
@@ -0,0 +1,130 @@
...
+            <SliderButton
+              slideId={index}
+              key={slide.animationName}
+              isActive={activeId === index}
+              onClick={this.handleSliderButtonClick}
+            />
+          ))}
+        </div>
+        <JLink href='/wallets' className={IntroductionStyle.link}>
+          <JRaisedButton
```
stcherenkov: When you put button inside link, it creates dangerous semantic that could work strange on some devices. Use `JLink` with button styles instead. For reference, see here: https://github.com/jibrelnetwork/jwallet-web/blob/741662f449fb139ba54a44969548ca73374d9da4/src/pages/WalletsStart/components/NewWalletButtons/NewWalletButtons.js#L39-L46

stcherenkov: And `JLink` here is wrong. You don't need to redirect user anywhere, because it is a pop-up screen. User's destination is important, we do not want to lose it by changing URL.

PR #776, opened 2019-04-26
JWWEB-735: Add search for assets list

stcherenkov:

PR #777, opened 2019-04-26
JWWEB-735: Add assets list for home page

stcherenkov:
Requires #776 and some integration work


stcherenkov:
Nicely done, thanks!

src/utils/address/startsWithOrEndsWith.js
```diff
@@ -0,0 +1,12 @@
+// @flow strict
+
+import { endsWith } from 'lodash-es'
+
+export function startsWithOrEndsWith(initialAddress: string, initialQuery: string): boolean {
+  const address = initialAddress.toLowerCase()
+  const query = initialQuery.toLowerCase()
+
+  return address.startsWith(query)
+    || address.startsWith(`0x${query}`)
```
IViol: isn't it possible to pass query, that already contain `0x`?
probably, need to use `add0x` util function here

Creepypoke: `.startsWith` [is not works](http://kangax.github.io/compat-table/es6/#test-String.prototype.startsWith) in Internet Explorer

src/pages/Home/HomeView.js
```diff
@@ -1,64 +1,157 @@
...
+        return (
+          <li key={address}>
+            <AssetItem address={address} />
+          </li>
+        )
+      })}
+    </ul>
+  )
+
+  renderEmptyList = () => (
```
stcherenkov: Separate component maybe?

src/utils/address/startsWithOrEndsWith.js
stcherenkov: Does babel transform it for IE?

src/pages/Home/HomeView.js
Creepypoke: yes, we make it in the next iteration

stcherenkov:
👍

PR #782, opened 2019-05-13
Remove unused and update existing babel deps

stcherenkov:


IViol:
Did you check building of `prod`?

stcherenkov:
> Did you check building of `prod`?

Checked it, works fine

PR #785, opened 2019-05-14
Input & textarea fixes

IViol:
- Update & align design of input & textarea fields
- Prepare textarea for using in forms


src/components/base/JTextArea/JTextArea.js
```diff
@@ -1,133 +1,182 @@
...
-    onFocus?: () => mixed,
-    label?: string,
-    id?: string,
-    value?: any, // In common case it will be string, but in React value of input has type `any`
-    rows?: number,
-    error?: boolean,
-    disabled?: boolean,
-  }
+type Props = StyleComponent<Theme> & {
+  +onChange: (?SyntheticEvent<HTMLTextAreaElement>) => void,
```
Creepypoke: if we suggesting that this property of record is not required, I would prefer use `onChange?: ...`, in case if we suggesting that property is required, but handler is not, I would prefer use `defaultProps` with value `noop`

PR #786, opened 2019-05-14
Normalize default browser styles

Creepypoke:


src/styles/mixins.scss
```diff
@@ -5,7 +5,7 @@
 }
 
 @mixin font($font-size: 13px, $line-height: 1) {
-  font-family: $font;
+  font-family: $font !important;
```
stcherenkov: There is no other way? I tried `font-family: inherit;`, works in my cases. Can you check it out?

Creepypoke: Yeah, `inherit` should work fine, but, from my point of view: we don't need to have a opportunity to change `font-family` anywhere. So, if it wrong, let's remove `!important`

PR #787, opened 2019-05-15
Update utils

IViol:
- Update export syntax
- Update for new create/import wallet changes

**Note**: `wallets` worker currently unused, please ignore its fixes.


src/utils/mnemonic/getPrivateHdRoot.js
```diff
@@ -1,18 +1,15 @@
-// @flow
+// @flow strict
 
 import bitcore from 'bitcore-lib'
 
-import getHdPath from './getHdPath'
+import { getHdPath } from '.'
```
stcherenkov: Maybe better to import from file directly? This will help to avoid potential circular dependencies

IViol: Take a look at the next example:

*src/moduleA/x.js*
```
import { y } from 'moduleB'
...
```

*src/moduleB/y.js*
```
import { x } from 'moduleA'
...
```

The same circular dependencies problem. It seems doesn't matter which way we use to import (file or module).
If so, I would suggest to conform our rule to use module everywhere (if it exists) for importing.

stcherenkov: Ok!

PR #798, opened 2019-05-17
Replaced values of properties to variables

Creepypoke:


src/styles/_base.scss
```diff
@@ -24,6 +24,7 @@ $font: 'Open Sans', Arial, sans-serif;
  * Page Header
  */
 $page-header-height: 130px;
+$new-page-header-height: 112px;
```
stcherenkov: Please add FIXME comment to remove old value and rename the new one, so we wouldn't end up with `new-new-new-` names

PR #801, opened 2019-05-17
Fix JCheckbox component

IViol:
**Note**: icons were not updated


src/components/base/JCheckbox/JCheckbox.js
```diff
@@ -33,48 +27,40 @@ class JCheckbox extends PureComponent<Props> {
...
+          className={jCheckboxStyles.field}
+        >
           <input
             onChange={this.handleChange}
-            name={`checkbox-${name}`}
+            name={name}
+            className={jCheckboxStyles.input}
             type='checkbox'
-            className='checkbox'
             defaultChecked={isChecked}
```
stcherenkov: Maybe it's better for input to be controlled?

src/components/base/JCheckbox/JCheckbox.js
```diff
@@ -1,29 +1,23 @@
...
-  JText, JIcon,
-} from 'components/base'
+import { JIcon } from 'components/base'
+
+import jCheckboxStyles from './jCheckbox.m.scss'
 
-type JCheckboxColor = 'white' | 'gray'
+type JCheckboxHandler = (boolean) => void
 
 type Props = {|
```
Creepypoke: I guess, checkbox is [Base Component](https://github.com/jibrelnetwork/frontend/blob/master/architecture/components.md#base-components), and make sense add `StyledComponent` type to Props

PR #808, opened 2019-05-20
JWWEB-730 Create wallet process refactoring

IViol:


src/components/WalletBackupForm/WalletBackupForm.js
```diff
@@ -16,7 +16,7 @@ import {
 import walletBackupFormStyle from './walletBackupForm.m.scss'
 
 type Props = {|
-  +handleSubmit: (?SyntheticEvent<HTMLFormElement>) => Promise<?FormFields>,
+  +handleSubmit: (?SyntheticEvent<HTMLFormElement>) => ?Promise<?FormFields>,
```
IViol: fixed because of `flow`

IViol: final-form handleSubmit has optional promise type for return value

src/pages/WalletsCreate/WalletsCreateView.js
```diff
@@ -1,122 +1,319 @@
...
   componentDidMount() {
-    this.props.openView()
+    this.props.requestBlockNumbers()
   }
 
-  componentWillUnmount() {
-    this.props.closeView()
+  setCurrentStep = (currentStep: WalletsCreateStep) => () => {
+    this.setState({ currentStep })
+    gaSendEvent('CreateWallet', EVENTS[currentStep])
```
IViol: need to fix

src/components/WalletBackupForm/WalletBackupForm.js
```diff
@@ -75,13 +76,12 @@ export function WalletBackupForm({
...
       <h2 className={walletBackupFormStyle.title}>{t`Back Up ${name}`}</h2>
       <p className={walletBackupFormStyle.text}>
-        {hasMnemonicAttributes
-          ? getMnemonicAttributesFirstLine(passphrase, derivationPath)
-          : BACKUP_KEY_FIRST_LINE
-        }
-      </p>
-      <p className={walletBackupFormStyle.text}>
-        {hasMnemonicAttributes ? BACKUP_MNEMONIC_SECOND_LINE : BACKUP_KEY_SECOND_LINE}
+        {backupText.map((text: string, index: number) => (
```
stcherenkov: Maybe `.split('\n').join('<br />')` and `dangerouslySetInnerHTML` would look simpler?

src/components/WalletBackupForm/WalletBackupForm.js
```diff
@@ -113,7 +113,7 @@ export function WalletBackupForm({
           type='button'
           theme='general'
           className={walletBackupFormStyle.button}
-          onClick={handleDownload}
+          onClick={handleDownload(values.data, passphrase, derivationPath)}
```
stcherenkov: Bad practice, you create new function on every props check resulting in rerender.

You need to memoize the resulting function or make it method of class instance.

IViol: we always have just one set of these props

stcherenkov: It does not matter. You still create a new function. And react always calls render (or function component body in your case) and then makes strict comparison of props by reference – and you will always have a new reference here.

Try in browser console, for example:
```js
function foo(i, j) {
  return () => i + j
}

console.log(foo(0, 1) === foo(0, 1))
```

Arguments are the same, but results are not equal.

IViol: Sure, I know it.
But especially for this component it absolutely is not a problem.

stcherenkov: True.

PR #813, opened 2019-05-21
Fix hover color for NewWalletButtons

IViol:


Creepypoke:
This is global problem. I'll fix it in other way, by removing modern-normalize.

stcherenkov:
Fixed in #819

PR #816, opened 2019-05-22
JWWEB-782 Wallet Item: start work

Dragonesis:


src/components/WalletFace/Actions/WalletFaceActions.js
```diff
@@ -71,16 +72,28 @@ class WalletFaceActions extends PureComponent<Props> {
...
 
     return (
-      <div className={classNames('wallet-face-actions', isToggled && '-toggled')}>
-        <div onClick={toggle} className='overlay' />
-        {isToggled ? (
-          <div className='actions'>
-            {balance && (
+      <div className={classNames(
+        walletFaceActions.core,
+        isToggled && walletFaceActions['-toggled'],
```
IViol: we stopped using of rscss namings, because we don't need modifiers with css modules (styles have plain structure).
try to edit old class like `variants` and `elements` from rscss.

src/components/WalletFace/Actions/walletFaceActions.m.scss
```diff
@@ -0,0 +1,57 @@
+@import 'styles/_base';
+@import 'styles/mixins';
+@import 'styles/colors';
```
IViol: mixins includes base which includes colors

src/components/WalletFace/Actions/walletFaceActions.m.scss
```diff
@@ -0,0 +1,57 @@
...
+
+.overlay {
+  z-index: 2;
```
IViol: As the `z-index` is "global" attribute (nodes with it affect other nodes) better to control all its values in one place.
We have that one - styles/zindex. We should keep it up to date.

src/components/WalletFace/walletFace.m.scss
```diff
@@ -0,0 +1,104 @@
...
+
+.actions {
+  @include hover-animation;
+  @include flex-column-center;
+
+  height: 100%;
+  flex: 0 0 56px;
+}
+
+.core {
```
IViol: better to move `.core` at the top

src/components/MnemonicAddresses/MnemonicAddresses.js
```diff
@@ -2,45 +2,38 @@
...
 
-import {
-  handle,
-  ignoreEvent,
-} from 'utils/eventHandlers'
-
 type Props = {|
-  +renameAddress: (Address) => void,
-  +setActive: (addressIndex: Index) => void,
+  // +renameAddress: (Address) => void,
```
stcherenkov: Why comments?

src/pages/Wallets/Wallets.js
```diff
@@ -0,0 +1,21 @@
...
+
+function mapStateToProps(state: AppState) {
+  return {
+    items: selectWalletsItems(state),
+    activeWalletId: selectActiveWalletIdOrThrow(state),
+  }
+}
+
+export const Wallets = (
+  connect/* :: < AppState, any, OwnPropsEmpty, _, _ > */(mapStateToProps)
```
stcherenkov: No need for comment now

stcherenkov: @flexxnn fixed flow to use connect types without commenting it

src/components/WalletCard/WalletCardActions/WalletCardActions.js
```diff
@@ -0,0 +1,159 @@
...
+          </JLink>
+          <JLink
+            href={`/wallets/${id}/addresses`}
+            className={walletCardActionsStyles.action}
+          >
+            {t`Manage Addresses`}
+          </JLink>
+          {isMultiAddressWallet && (
+            <JLink
+              href={`/wallets/${id}/mode`}
```
stcherenkov: Better to make this a state change (internal step), not link. We don't want user to save this link then try to figure out what happens.

PR #823, opened 2019-05-23
Fix JInputField component styles

flexxnn:


src/components/base/JInputField/JInputField.js
```diff
@@ -71,6 +70,7 @@ export function JInputField({
       )}
     >
       <div
+        onClick={handleFocus(textInput)}
```
stcherenkov: Creates new function on every call to render. It will cause continuous re-renders, that could lead to performance issues or memory leaks. Please use class method, memoization or hook instead.

src/components/base/JInputField/jInputField.m.scss
```diff
@@ -15,12 +15,7 @@
 .wrap {
   height: 72px;
   padding: 0;
-  box-shadow: $normal-shadow;
```
stcherenkov: All inputs have shadow in design. Why did you remove it?

flexxnn: This was not removed, it was moved to the color theme section, because we need to hide shadow in the disabled state
```
.white,
.white-indicator {
  .wrap {
    background-color: $white;	    background-color: $white;
    box-shadow: $normal-shadow;

     &:hover {
      box-shadow: $hover-shadow;
    }
  }
...
```

src/components/base/JInputField/jInputField.m.scss
```diff
@@ -93,35 +86,40 @@
...
+    background: $white;
+
+    &:hover {
+      box-shadow: none;
+    }
 
     .label {
-      user-select: none;
       cursor: default;
+      color: $main-black-40;
```
IViol: but it must be `$main-black-30`

IViol: please recheck figma screens

flexxnn: Checked and fixed
value + disabled - 30%
just disabled with value - 40%

PR #828, opened 2019-05-27
Add margin-top 16px for Button component

IViol:
Designers updated their sketches with additional margin ahead of buttons. I checked figma screens, and found one place, where we have just 8px between buttons. I think it's ok to manage margin-top for button from one place - its css file to avoid adding it for each form. What you think?

### DIFF
--- a/src/components/base/Button/button.m.scss
+++ b/src/components/base/Button/button.m.scss
@@ -10,6 +10,7 @@
   height: 48px;
   width: 100%;
   padding: 0 25px;
+  margin-top: 16px;
   border-radius: 4px;
   border: none;
   cursor: pointer;


Creepypoke:
First of all, I disagree that base components must make guess the context where they will be placed, even we have equal margins between buttons and other elements in design, it's question of responsibility separation between components. Second point is: we have defined [offsetStyles](https://github.com/jibrelnetwork/jwallet-web/blob/develop/src/styles/offsets.m.scss), I think, they need to set margins more straightforward and explicity.

IViol:
@stcherenkov @flexxnn what you think?

stcherenkov:
I like offsetStyles more, because it's both transparent and reusable. And also margin act kind of unpredictable because of [margin collapsing](https://css-tricks.com/what-you-should-know-about-collapsing-margins/). It's better to have bottom margin everywhere or top margin everywhere, not both.

PR #830, opened 2019-05-27
JWWEB-802 Fix text field design issues

IViol:


src/components/base/JInputField/JInputField.js
```diff
@@ -71,6 +70,7 @@ export function JInputField({
       )}
     >
       <div
+        onClick={handleFocus(textInput)}
```
stcherenkov: We need to fix this some day

PR #833, opened 2019-05-27
Add standard decorator function to validate all required fields

stcherenkov:
## What is it for?

To simplify implementation of form behavior discussed with design and described [in Confluence](https://jibrelnetwork.atlassian.net/wiki/spaces/JWALLET/pages/830504982/UX+Patterns) (see Forms Validation).

In short: we don't show field errors for required fields, but we also disable primary button while there are required fields that are empty.

## What does it do?

It creates one more error for you if one of required fields is not specified. This error can be used to disable submit button (and you can't submit final form while you have errors in it).

## How does one use it?

You just compose it with your validation function, like this:

```js
const composedValidate = finalFormComposeWithValidateRequired({
  requiredKeys: ['foo', 'bar'],
}, (values) => {
  // your function here
})
```

Then pass it to Final Form:

```js
<Form
  onSubmit={this.handleSubmit}
  validate={composedValidate}
  render={this.renderForm}
  initialValues={INITIAL_VALUES}
/>
```

Your function will run first, then required keys will be checked for existence.

For more details see tests.


stcherenkov:
Simplified it a bit.

Now we have 2 functions.

First – composes multiple validators (both sync and async) into one.

Second – checks if required fields are not empty.

So if you want to check all fields for non-empty values, and then run your own checks, you should do it like this:

```js
const composedValidate = allValidators([
  validateRequired(),
  (values) => {
    // your function here
  },
])
```

So with this update we can make additional utility functions for validations and compose them all together.

PR #835, opened 2019-05-28
Minor import fixes

IViol:


src/components/WalletPasswordForm/walletPasswordForm.m.scss
```diff
@@ -5,5 +5,11 @@
 
   position: relative;
   width: 544px;
+  height: 100%;
   margin: 0 auto;
 }
+
+.forgot {
+  position: absolute;
```
stcherenkov: Did you check it on mobile?

src/pages/WalletsImport/walletsImport.m.scss
```diff
@@ -1,7 +1,10 @@
 @import 'styles/mixins';
 
 .core {
+  @include flex-row-center;
+
   width: 100%;
+  height: 100%;
```
stcherenkov: Looks ok on narrow screens?

src/components/WalletPasswordForm/walletPasswordForm.m.scss
IViol: just in chrome devices emulator

src/pages/WalletsImport/walletsImport.m.scss
IViol: Is there any place with etalon sketches for narrow screens? Which we can compare with our results?

stcherenkov: No, we just check it for sanity right now. If it doesn't look bad or broken, it's fine

PR #841, opened 2019-05-29
JWWEB-737 Fixed visual issues, fix formatter of address and amount

Creepypoke:


src/components/HistoryItemDetails/HistoryItemDetailsInternal.js
```diff
@@ -37,8 +37,14 @@ export class HistoryItemDetailsInternal extends PureComponent<Props> {
       return null
     }
 
+    const extendedProps = {
+      ...this.props,
+      editNote: this.debouncedHandlerOnEditNote,
+      key: this.props.id,
+    }
```
IViol: Does flow able to check such structures?
Probably better to define specific type for it.

Creepypoke: haha! Already done, but not pushed to repo. I'm playing with it right now :)

src/components/TransactionsList/TransactionsList.js
```diff
@@ -44,26 +45,20 @@ class TransactionsList extends PureComponent<Props, State> {
     this.rootWrapper.addEventListener('scroll', this.handleScroll)
```
IViol: why don't just `window`?

Creepypoke: Because scroll in our app set in [`#root` element](https://github.com/jibrelnetwork/jwallet-web/blob/develop/src/styles/core.scss#L23), and scroll of window don't handle.

src/store/utils/HistoryItem/types.js
```diff
@@ -0,0 +1,71 @@
...
+  +type: typeof EVENT_BURN_TYPE,
+|}
+
+export type HistoryItem
+  = TransferIn
+  | TransferOut
+  | TransferCancel
+  | ContractCall
+  | EventMint
+  | EventBurn
```
IViol: I would suggest to move this file into `flow-typed/redux/history`, and `declare` many of these types.

src/store/utils/HistoryItem/utils/getTransactionFee.js
```diff
@@ -0,0 +1,16 @@
...
+  transaction: TransactionWithPrimaryKeys,
+  asset: DigitalAsset,
+): string {
+  const gasUsed = get(transaction, 'receiptData.gasUsed', 0)
+  const gasPrice = get(transaction, 'data.gasPrice', 0)
+  const decimals = get(asset, 'blockchainParams.decimals', 18)
+
+  return getTxFee(gasUsed, gasPrice, decimals)
+}
+
```
IViol: btw, we already have similar util - https://github.com/jibrelnetwork/jwallet-web/blob/develop/src/utils/transactions/getTxFee.js

src/store/utils/HistoryItem/types.js
Creepypoke: Global declarations of types is bad practice, because  they don't declare a context of using

PR #844, opened 2019-05-30
Add proxying of onClick event for JIcon

IViol:


IViol:
ok, your point is fair enough. Im closing this

PR #845, opened 2019-05-30
Update CopyableField component

IViol:


Creepypoke:
I'm sure, that you can get same functionality from [this component](http://app.jwallet.develop.stack.jdev.network/storybook/?path=/story/fieldpreview--default)

IViol:
`FieldPreviewInternal` component has unsuitable (as well as too complex) API and unsuitable implementation. For example, it has `copy` prop, which is verb, so it should be some handler. This component shouldn't control which value will has alert to show message for user, it only must call action to initiate displaying of alert. Its API complex also because it combines three independent actions - copy to clipboard, add to user's contacts list, and render link.
I found that easiest way for me is to write new correct component, that has minimal API and implementation and don't break all our codestyle rules.

PR #846, opened 2019-05-30
Update WalletBackupForm

IViol:


src/components/WalletBackupForm/WalletBackupForm.js
```diff
@@ -2,22 +2,20 @@
 
 import React, { PureComponent } from 'react'
 import { t } from 'ttag'
-import { Field } from 'react-final-form'
 
 import { fileSaver } from 'services'
+import { CopyableField } from 'components'
```
Creepypoke: We already have [`FieldPreview`](http://app.jwallet.develop.stack.jdev.network/storybook/?path=/story/fieldpreview--default) for the same view

PR #849, opened 2019-05-30
JWWEB-816 Update wallet backup screen

IViol:


src/pages/WalletsItemBackup/WalletsItemBackupView.js
```diff
@@ -0,0 +1,265 @@
...
+              }
+
+              const passphrase: string = decryptData({
+                key: internalKeyDec,
+                data: encrypted.passphrase,
+              })
+
+              this.setState({ data })
+
+              this.setState({
```
Creepypoke: It's work as designed: two asynchronous `setState` calls one by one?

IViol: it will be fixed

PR #850, opened 2019-05-31
Fix handleFocus for text fields

IViol:


src/components/base/JInputField/JInputField.js
```diff
@@ -81,7 +85,6 @@ export class JInputField extends PureComponent<JInputFieldProps> {
 
     return (
       <div
-        onClick={this.handleFocus}
```
IViol: it will be called by click on inner child

PR #857, opened 2019-06-04
JWWEB-737 Added nonce to transaction data from node

Creepypoke:
Also type `Transaction` was merged to `TransactionWithPrimaryKeys` via spread.

# What to look for
I'm not sure about `nonce` for `TransferEventFromEthereumNode` and `JNTEventFromEthereumNode`, because it's property of their "parent" transaction entity, and I could not understand what actually we get to their handle functions 🥒


flow-typed/redux/transactions.js
```diff
@@ -1,10 +1,12 @@
 // @flow
 
 declare type Hash = string
+// TransactionId consist of: `${txHash}${?txLogIndex}`
 declare type TransactionId = string
 
 declare type TransactionFromBlockExplorer = {|
   +hash: Hash,
+  +nonce: string,
```
IViol: To avoid parsing of string I suggest to make it as `number` type.
Type of that param in response is number

flow-typed/redux/transactions.js
```diff
@@ -119,23 +124,12 @@ declare type Transaction = {|
 |}
 
 declare type TransactionWithPrimaryKeys = {|
+  ...Transaction,
```
IViol: `...$Exact<Transaction>,`

Creepypoke: `Transaction` exact structure already https://flow.org/en/docs/types/utilities/#toc-exact

flow-typed/redux/transactions.js
Creepypoke: But value from node is `string`

PR #861, opened 2019-06-05
Code cleanup

IViol:


src/store/sagas/digitalAssets/send.js
```diff
@@ -576,7 +576,7 @@ function* sendTransactionRequest(
   yield put(digitalAssetsSend.setIsLoading(true))
 
   try {
-    const privateKey: string = yield* getPrivateKey(walletId, password)
+    const privateKey: string = '' // yield* getPrivateKey(walletId, password)
```
Creepypoke: ? :)

src/store/sagas/digitalAssets/send.js
```diff
@@ -25,8 +25,8 @@ import {
 } from 'utils/address'
 
 import {
-  getPrivateKey,
-  GetPrivateKeyError,
+// getPrivateKey,
```
Creepypoke: Why just remove this code too?

IViol: This is dead code. Comments will be completely removed during refactoring send asset process.
Actual code for getting of privateKey is https://github.com/jibrelnetwork/jwallet-web/blob/develop/src/store/plugins/walletsPlugin.js#L524

PR #864, opened 2019-06-07
JWWEB-791 Wallets item upgrade

IViol:


src/store/plugins/walletsPlugin.js
```diff
@@ -77,7 +77,7 @@ class WalletsPlugin {
 
   getItems = (): Wallets => selectWalletsItems(this.getState())
 
-  getInternalKey = async (password: string): Promise<?Uint8Array> => {
+  getInternalKey = async (password: ?string): Promise<?Uint8Array> => {
```
Creepypoke: We are have check about `maybe` value?

PR #874, opened 2019-06-24
JWWEB-856 Markup for Contact Page

Creepypoke:
Markup only

Access to page via url: `/contacts/:any-text`


flow-typed/types.js
```diff
@@ -10,6 +10,7 @@ declare type BalanceString = string
 
 declare type Address = string
 declare type OwnerAddress = Address
+declare type ContactId = OwnerAddress // It's just address yet
```
IViol: there is specific file `flow-typed/redux/favorites`. I suggest rename it to `contacts` and place new types into it.

Creepypoke: I have heard that we don't plan to store contacts in redux storage

IViol: ...oneday
:D

PR #879, opened 2019-06-24
JWWEB-789, JWWEB-790: Wallets item mode pages

IViol:


src/pages/WalletsItemModeDisable/WalletsItemModeDisable.js
```diff
@@ -0,0 +1,43 @@
...
+import { walletsPlugin } from 'store/plugins'
+
+import walletsItemModeDisableStyle from './walletsItemModeDisable.m.scss'
+
+export type Props = {|
+  +walletId: string,
+|}
+
+export class WalletsItemModeDisable extends PureComponent<Props> {
+  handleDisable = () => {
```
Creepypoke: If it would be separated function instead method of class, then component would be stateless and we could make it memoized via `React.memo`.

src/pages/WalletsItemModeEnable/WalletsItemModeEnableView.js
```diff
@@ -0,0 +1,48 @@
...
+} from 'components'
+
+import walletsItemModeEnableStyle from './walletsItemModeEnable.m.scss'
+
+export type Props = {|
+  +goBackToWallets: () => any,
+  +walletId: string,
+|}
+
+export class WalletsItemModeEnableView extends PureComponent<Props> {
```
Creepypoke: It's looks very similar to `WalletsItemModeDisable`, maybe make sense to extract some common things to component?

src/pages/WalletsItemModeDisable/WalletsItemModeDisable.js
IViol: Difficult to imagine situation when this page will be re-rendered several time for user. So I think we will not have any benefits with memo.

src/pages/WalletsItemModeEnable/WalletsItemModeEnableView.js
IViol: For example?

IViol: I think that all their common things I reused via `UserActionInfo` component.

PR #880, opened 2019-06-24
Language picker

flexxnn:


src/data/settings/languages.js
```diff
@@ -1,10 +1,78 @@
...
+    title: 'Japan',
+    code: 'ja',
+    iconName: 'ic_japan_24',
+    isDisabled: false,
+  },
+  {
+    // TODO: translate to Spanish
+    title: 'Spanish',
+    code: 'es',
+    iconName: 'ic_spain_24',
```
IViol: I suggest to remove `ic_` & `_24` common parts

IViol: but would be great if we will use computed iconName, like `ic_${code.toLowerCase()}_24`

PR #881, opened 2019-06-25
JWWEB-860 Added page for deleting contact item

Creepypoke:
This PR use another PR #882, check it first, please


src/pages/ContactsItemDelete/ContactsItemDelete.js
```diff
@@ -0,0 +1,24 @@
...
+  type Props,
+  ContactsItemDeleteView,
+} from './ContactsItemDeleteView'
+
+type OwnProps = {|
+  +contactId: ContactId,
+|}
+
+function mapStateToProps(state: AppState, { contactId }: OwnProps) {
+  const name = 'Sarah Jessica Parcel'
```
flexxnn: I think we need TODO comment, where we hardcode something

PR #883, opened 2019-06-25
JWWEB-783 Update wallets page

IViol:


src/components/TitleHeader/TitleHeader.js
```diff
@@ -4,34 +4,42 @@ import React from 'react'
...
       {handleClick && (
         <div
           onClick={handleClick}
-          className={titleHeaderStyle.back}
+          className={style.back}
         >
           <JIcon name='arrow-back-use-fill' size='medium' color='blue' />
         </div>
       )}
+      <h1 className={style.title}>{title}</h1>
```
Creepypoke: When `children` not define this block steal has `margin-bottom`, is it work as designed?

IViol: Yes, its `margin-bottom` doesn't affect `aside`, because container is row.

PR #884, opened 2019-06-25
JWWEB-858 Added add contact page

Creepypoke:


src/pages/ContactsItemAdd/ContactsItemAdd.js
```diff
@@ -0,0 +1,36 @@
...
+  state: AppState,
+  {
+    address,
+    name,
+    note,
+  }: OwnProps,
+) {
+  const handledAddress = address || ''
+  const handledNote = note || ''
+  const handledName = name || 'Sara Jessica Parcel'
```
flexxnn: ????
'Sara Jessica Parcel'

:)

Creepypoke: I prepare the project to best future :D

src/pages/ContactsItemAdd/ContactsItemAddView.js
```diff
@@ -0,0 +1,106 @@
...
+              label={t`Name`}
+              name='name'
+              isDisabled={submitting}
+            />
+            <Field
+              component={JInputField}
+              label={t`Address`}
+              name='address'
+              errorMessage={errors.address}
+              isDisabled={submitting || address}
```
IViol: isDisabled is true when address exist and not empty. Is it correct?

Creepypoke: Yeah, because address is not editable id in this case.

PR #891, opened 2019-06-26
JWWEB-855 Page of Contacts list

Creepypoke:
Draft
Based on #890


src/components/ContactItem/ContactItem.js
```diff
@@ -25,9 +25,18 @@ type Props = {|
   ...OwnProps,
 |}
 
-function Component(props: Props) {
-  const splitedName = props.name.split(' ')
+function getName(dividedName: string[]) {
```
flexxnn: Uppercase

flexxnn: I suggest to move it to utils

PR #900, opened 2019-06-28
Fix additional button vertical alignment

Creepypoke:
Fixed visual issues of additional-button


IViol:
Please check develop. Seems that this problem has been already resolved.

PR #901, opened 2019-07-01
JWWEB-870 Wallet address rename

IViol:


src/pages/WalletsItemAddresses/components/WalletAddressCard/WalletAddressCard.js
```diff
@@ -27,18 +30,26 @@ type Props = {|
...
 type StateProps = {|
+  +newName: string,
   +ethBalance: ?BigNumber,
+  +isRenameActive: boolean,
 |}
 
 const MINUTE: number = 60 * 1000
 
 class WalletAddressCard extends Component<Props, StateProps> {
+  nameInputRef: Object
```
flexxnn: #fixme

```
type NameInputRef = {
  current: null | HTMLInputElement,
}

...

nameInputRef: NameInputRef 
```

src/pages/WalletsItemAddresses/components/WalletAddressCard/WalletAddressCard.js
```diff
@@ -66,13 +77,49 @@ class WalletAddressCard extends Component<Props, StateProps> {
...
+
+  handleRenameBlur = () => {
+    this.handleActivateRename(false)
+
+    const {
+      address,
+      addressName,
+    } = this.props
+
+    const newName: string = this.state.newName.substring(0, 32).trim().replace(/\//g, '–')
```
flexxnn: May be change order: `trim->replace->substring(...)`. Sometimes, this will allow to use strings with bigger length

src/pages/WalletsItemAddresses/components/WalletAddressCard/WalletAddressCard.js
IViol: fixed in b241ba8

src/pages/WalletsItemAddresses/components/WalletAddressCard/WalletAddressCard.js
IViol: fixed in 374e4df

PR #904, opened 2019-07-01
Update wallet rename

IViol:


src/components/WalletCard/WalletCard.js
```diff
@@ -25,15 +26,21 @@ type Props = {|
 
 type StateProps = {|
   +newName: string,
+  +isNewNameUniq: boolean,
   +isRenameActive: boolean,
 |}
 
+function sanitizeName(name: string) {
+  return name.substring(0, 32).trim().replace(/\//g, '–')
```
flexxnn: May be, replace then trim will be better?

flexxnn: Do we need to move this fn to utils to share it?

IViol: > May be, replace then trim will be better?

it doesn't matter, because there are no any spaces after `replace`

IViol: > Do we need to move this fn to utils to share it?

these actions are actual during renaming of a wallet only

IViol: fixed in 374e4df

PR #906, opened 2019-07-02
Update wallet item

IViol:


src/components/WalletCard/WalletCard.js
```diff
@@ -113,30 +138,27 @@ class WalletCard extends Component<Props, StateProps> {
...
       isNewNameUniq,
       isRenameActive,
     }: StateProps = this.state
 
-    const addressesCount: number = 5
     const isMultiAddress: boolean = !!xpub
-    const addressName: ?string = 'Address Name'
     const isActive: boolean = (id === activeWalletId)
     const hasMessage: boolean = (!isNewNameUniq && isRenameActive)
+    const name: string = addressName ? `${addressName}  •  ` : ''
```
flexxnn: We have `src/utils/getAddressName` for get address name for multi address wallets

src/components/WalletCard/WalletCard.js
```diff
@@ -217,19 +241,36 @@ class WalletCard extends Component<Props, StateProps> {
...
+  const {
+    name,
+    xpub,
+    customType,
+    derivationIndex,
+    isSimplified,
+  }: Wallet = wallet
+
+  const addressName: ?string = !isSimplified
+    ? addressNames[address] || t`Address ${addessIndex}`
```
flexxnn: `utils/address/getAddressName`

src/pages/WalletsItemAddresses/WalletsItemAddressesView.js
```diff
@@ -27,21 +28,45 @@ export type Props = {|
...
       addresses: this.deriveAddresses(props),
     }
   }
 
-  componentWillReceiveProps(nextProps: Props) {
+  async componentDidMount() {
+    this.requestETHBalance()
+  }
+
+  async componentWillReceiveProps(nextProps: Props) {
```
flexxnn: This is deprecated method. We should not use it, or use `UNSAFE_` prefix if there is no way to make it.

https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

src/store/plugins/walletsPlugin.js
```diff
@@ -411,6 +419,47 @@ class WalletsPlugin {
...
     )
   }
+
+  requestETHBalanceByAddress = async (address: Address): Promise<string> => {
+    const network: Network = this.getNetwork()
+
+    try {
+      const ethBalance: string = await web3.getAssetBalance(network, address, 'Ethereum')
+
+      return new BigNumber(ethBalance)
```
flexxnn: This will re-request balance when server answers with invalid value and it can not be converted to BigNumber.
I think, it need to re-request data only when network error occurred. When server answers with 200 responce, but value is invalid we should set error for this balance value

src/store/plugins/walletsPlugin.js
```diff
@@ -411,6 +419,47 @@ class WalletsPlugin {
...
+    }
+  }
+
+  requestETHBalanceByXPUB = async (
+    walletId: WalletId,
+    derivationIndex: number,
+  ): Promise<BigNumber> => {
+    const addresses: Address[] = this.getAddresses(walletId, 0, derivationIndex)
+
+    return Promise
```
flexxnn: Promise.map is not standard promise API
Can be replaced by
```
const foo = addresses.map(address => this.requestETHBalanceByAddres(address))
Primise.all(foo).reduce(...
```

Not need to fix now, if works good

src/store/plugins/walletsPlugin.js
IViol: Do you know case when answer code is 200 and value is not valid bignumber?

PR #907, opened 2019-07-02
JWWEB-855 Fixed import of contacts list

Creepypoke:
Contacts list from json was replaced to js notation


IViol:
why?

Creepypoke:
JSON don't work correctly in production build

PR #924, opened 2019-07-05
Add toasts implementation

IViol:
To show toast, need to `dispatch` [showToast](https://github.com/jibrelnetwork/jwallet-web/blob/659483491d766274409d8b58f8fa7a8b1b82de8a/src/store/modules/toasts.js#L6-L11) action with [payload](https://github.com/jibrelnetwork/jwallet-web/blob/feature/toasts/flow-typed/redux/toasts.js#L22-L27) - [example](https://github.com/jibrelnetwork/jwallet-web/blob/feature/toasts/src/components/CopyIconButton/CopyIconButton.js#L36-L39)

Example of working with appearance animation you can check at Manage Addresses page for any multi-address wallet:
Screenshot 2019-07-05 at 13 32 55


IViol:
> Nope, toast display function should be abstract from store.
there are few ways to implement toasts:
- with the help of redux (ok)
- wrap entire AppContainer with some ToastContainer (shit)
- direct DOM changing (bullshit)

So, nope, toast display function should NOT be abstract from store

> We need a simple function to call
Look how I rewrite it with store plugin

> We don't need complex API either
your suggestion is ok, it is just PoC

IViol:
I left "options" (second argument) in toastPlugin API, it is currently unused, but could be helpful in future.
Now, `showToast` could be used in very simple way:

https://github.com/jibrelnetwork/jwallet-web/blob/16a9f69cfa94077c9b77c52e6ac045eb2f55d77d/src/components/CopyIconButton/CopyIconButton.js#L9

And then

https://github.com/jibrelnetwork/jwallet-web/blob/16a9f69cfa94077c9b77c52e6ac045eb2f55d77d/src/components/CopyIconButton/CopyIconButton.js#L32

That's all

PR #927, opened 2019-07-05
Fix bugs in Send Asset

flexxnn:


src/pages/Send/components/SendAmountField/sendAmountField.m.scss
```diff
@@ -134,6 +135,10 @@
   > .amount {
     color: $main-black-40;
     min-width: 60px;
+    text-overflow: ellipsis;
```
stcherenkov: Btw we have mixin for ellipsis

PR #930, opened 2019-07-08
Clean send asset

flexxnn:
Old files deleted


src/middlewares/analyticsMiddleware/index.js
```diff
@@ -7,7 +7,7 @@ import { resyncTransactionsEvents } from './resyncTransactionsEvents'
 export const analyticsMiddleware = store => next => (action) => {
   try {
     const startState = store.getState()
-    sendTransactionEvents(startState, action)
+    // sendTransactionEvents(startState, action)
```
stcherenkov: You can remove this line, everything is remembered in git history

PR #940, opened 2019-07-18
 JWWEB-875 - Edit contact

flexxnn:


src/pages/ContactsItemEdit/components/ContactEditForm/ContactEditForm.js
```diff
@@ -0,0 +1,136 @@
...
+  description: string, // note
+  name: string,
+|}
+
+export type Props = {|
+  +initialValues: FormValues,
+  +i18n: I18nType,
+  +checkContactExists: (address: OwnerAddress) => boolean,
+  +editContact: (contact: FormValues) => any,
+  +onDeleteClick: (address: OwnerAddress) => any,
```
IViol: I suggest to name handlers in the same way, e.g. editContact/deleteContact or onEdit/onDelete, etc

src/store/router/routes.js
```diff
@@ -60,15 +60,15 @@ export const routes: Array<{|
   name: 'ContactsItem',
   hasMenu: true,
 }, {
-  path: '/contacts/add?:address&:name&:note',
+  path: '/contacts/add?:address&:name',
```
IViol: why did you remove `note` param?

src/pages/ContactsItemEdit/ContactsItemEdit.js
```diff
@@ -1,26 +1,103 @@
...
+
+function mapStateToProps(
+  state: AppState,
+  { contactId }: OwnProps,
+) {
+  const {
+    address,
+    name = '',
+    description = '',
+  } = selectFavorite(state, contactId || '') || {}
```
IViol: Probably, better to `throw` if contactId is empty (actually, it is not possible, because of route's path - `/contacts/edit/:contactId`)

src/pages/ContactsItemEdit/components/ContactEditForm/ContactEditForm.js
stcherenkov: I don't agree. There is strong difference between edit and click on edit button. `onEdit` could mean: on click on edit button, on type in input, on change input, on focus in input and so on.

stcherenkov: I also suggest we put handler first, description second. So it would all be structured the same way for all the handlers, like: `onClick` (if it is the only element), `onClickEdit`, `onClickDelete`, `onChangeName`, `onFocusIcon` etc.

PR #948, opened 2019-07-25
Update transaction structure

IViol:
- add `hasInput` field into `Transaction` structure. It is needed to define "abstract" contract method call.
- remove redundant `isCanceled` flag from `Transaction` structure.
- add "filtering" flags into `TransactionsState`.
- fix validation of transactions from block explorer. Add ability to show "abstract" contract method calls.
- fix **critical** bug connected with incorrect `nonce` validation from web3 payload, that caused infinite ETH node syncing.
- update blocks & transactions selectors. Add ability to omit some optional parameters, that always defined in the app, such as network & currently active wallet address.
- update `selectTransactionById` / `selectPendingTransactionByHash`. These selectors was made flexible, added ability to search items with different precision (passing different set of arguments)


flow-typed/redux/transactions.js
```diff
@@ -8,6 +8,7 @@ declare type TransactionFromBlockExplorer = {|
   +hash: Hash,
   +to: string,
   +from: string,
+  +input: string,
```
flexxnn: I think we need to leave comment here. When this PR will be closed nobody will know what is this

src/services/blockExplorer.js
```diff
@@ -86,34 +81,37 @@ function checkETHTransaction(data: Object): boolean {
     type.isString(data.gas) &&
     type.isString(data.from) &&
     type.isString(data.hash) &&
+    type.isString(data.input) &&
+    type.isString(data.nonce) &&
```
flexxnn: As I remember, that getNonce method returns number. Please check that this is always a string

src/services/blockExplorer.js
```diff
@@ -86,34 +81,37 @@ function checkETHTransaction(data: Object): boolean {
...
 
     const {
+      input,
       value,
       isError,
       contractAddress,
     }: Object = item
 
     const isEmptyAmount: boolean = isZero(value)
+    const isEmptyInput: boolean = (input === '0x')
```
flexxnn: `input === '0x' || input === ''` ?

src/services/blockExplorer.js
IViol: I checked, etherscan API returns all fields as strings

src/services/blockExplorer.js
IViol: I checked, etherscan API returns `0x` string as empty input

flow-typed/redux/transactions.js
IViol: This field's name does not differ from ETH node getTransaction response format.
But, ok, will add some note in next PR

PR #957, opened 2019-07-26
 JWWEB-780 - Receive asset

flexxnn:


src/pages/DigitalAssets/index.js
```diff
@@ -4,7 +4,6 @@ import MenuLayout from 'layouts/MenuLayout'
```
IViol: this file is for react-router and need to be removed

src/pages/ReceiveAsset/ReceiveAsset.js
```diff
@@ -0,0 +1,157 @@
...
+    super(props)
+
+    this.state = {
+      isCopied: false,
+    }
+  }
+
+  componentDidMount() {
+    if (this.props.address) {
+      generateQRCode(this.props.address)
```
IViol: generateQRCode checked `if (address)...` twice

PR #975, opened 2019-08-06
Fix search

IViol:


src/utils/transactions/prepareListForRendering.js
```diff
@@ -17,13 +17,13 @@ const RE_INVALID_HEX: RegExp = /[^a-f0-9]/i
 
 const SEARCH_TRANSACTIONS_RULES: FilterPredicateRules = {
```
stcherenkov: Maybe it would be more reliable to use predefined functions instead of strings?

PR #996, opened 2019-08-16
JWWEB-862 Add info about asset

IViol:


src/pages/AssetsItem/components/About/components/Links/Links.js
```diff
@@ -0,0 +1,81 @@
...
+    <ul className={styles.core}>
+      {assetPage.urls.map((item: DigitalAssetPageUrl) => (
+        <li
+          key={item.url}
+          className={styles.link}
+        >
+          <JLink
+            href={item.url}
+            className={styles.anchor}
+            target='_blank'
```
stcherenkov: `JLink` sets these attributes to every external link by default.

https://github.com/jibrelnetwork/jwallet-web/blob/develop/src/components/base/JLink/JLink.js#L65-L74

I think, we should actually separate JLink into multiple components. Something like:
- RouterLink – for internal links and accept routeName (and proxy other Route props), not path
- ExternalLink – for external links with href and `_blank noopener noreferrer` by default
- TrustedLink – for external links we trust (like privacy policy, or maybe Jwallet landing)

The latter may be not needed, actually.

And then we share styles between them.

I think, this will remove a lot of confusion.

stcherenkov: Not in this PR, of course.

PR #997, opened 2019-08-16
Make HTTP listen port configurable

soar:
- Add Dockerize
- Move checks to an external Jenkins CI step
- Default HTTP port is set to 80


soar:
The master branch is broken, so it's unsafe to merge this. Let's move this work into develop branch.

PR #1001, opened 2019-08-20
Fix getting of env vars

IViol:


src/utils/config/getENVVar.js
```diff
@@ -1,4 +1,4 @@
-// @flow
+// @flow strict
 
 function getENVVar(key: string): ?string {
```
stcherenkov: We could do `getEnvVarOrDefault` here, passing default value as last argument, because this function is mostly used with default value.

IViol: I agree. But it will not work with current solution:
https://github.com/jibrelnetwork/jwallet-web/blob/develop/src/public/js/settings.js#L7-L8

PR #1004, opened 2019-08-22
JWWEB-952 Update showing of wallet/address names

IViol:


src/data/lingui/en/messages.json
```diff
@@ -407,5 +406,7 @@
     "AssetsItem.About.title": "About",
     "AssetsItem.Transfers.title": "Transfers",
     "HistoryList.Item.title.cancel": "Cancel transfer",
-    "AssetsItem.About.Description.less": "Show less"
+    "AssetsItem.About.Description.less": "Show less",
+    "utils.address.getAddressName.address": "Address {index}",
```
stcherenkov: Context is needed only to tell translator, where is this string used. That's why we created "entities" namespace – it contains strings related to specific business-entity. Why did you change it to `utils`?

src/data/lingui/en/messages.json
```diff
@@ -407,5 +406,7 @@
...
+    "utils.address.getAddressName.wallet": "{walletName} / {addressName}"
```
stcherenkov: This is technical string, that is not needed in translation.

src/data/lingui/en/messages.json
```diff
@@ -407,5 +401,11 @@
...
     "AssetsItem.Transfers.title": "Transfers",
     "HistoryList.Item.title.cancel": "Cancel transfer",
-    "AssetsItem.About.Description.less": "Show less"
+    "AssetsItem.About.Description.less": "Show less",
+    "SetPassword.hint.description": "If you forget your Security Password, some functions won’t be available. To restore access to all functions you will need to clear your data and re-import your wallets again using backup phrase.",
+    "SetPassword.hint.label": "Enter Password Hint",
+    "SetPassword.securityPassword": "Enter Security Password",
+    "SetPassword.submit": "Set Security Password",
+    "SetPassword.title": "Set Password to Secure Your Storage",
+    "entity.address.getAddressName.default": "Address {index}"
```
stcherenkov: Better to return address name to its previous state, because it is generally the same, but already translated.

PR #1016, opened 2019-09-10
Disable developer mode card in settings

IViol:


src/pages/Settings/components/Card/card.m.scss
```diff
@@ -18,6 +18,16 @@
   &:active {
     box-shadow: $select-shadow;
   }
+
+  &.disabled {
+    opacity: 0.5;
+    cursor: default;
```
stcherenkov: Maybe cursor should be defined in upper scope (in wrapping link, for example). I think, I read a couple of articles some time ago against redefining cursor with some performance issues, but I'm not sure and it was quite a while.

Anyway, it could be misleading – the item changes cursor, but is not clickable.

Not necessary to fix it in this PR though – just think about it.

PR #1025, opened 2019-09-13
Implement sorting of assets

IViol:


src/utils/digitalAssets/sortAssets.js
```diff
@@ -0,0 +1,53 @@
+// @flow strict
+
+import { checkETH } from '.'
+
+export function sortAssets(items: DigitalAssetWithBalance[]): DigitalAssetWithBalance[] {
```
stcherenkov: 👍Clean and neat, great work!

To make it even better, you can separate comparator function (it's much simpler to test, for example) and just call `list.sort(comparator)`.

PR #1037, opened 2019-09-24
JWWEB-891 Show wallet balance in fiat

IViol:


src/services/ticker.js
```diff
@@ -11,30 +11,43 @@ type TickerAPIParams = {|
...
 
-function callApi(params: TickerAPIParams, retryCount: number = 4): Promise<any> {
-  const requestInfo: RequestInfo = `${TICKER_API}/v2/quotes`
+function handleRequestError(
+  params: TickerAPIParams,
+  retryCount: number,
+) {
+  if (retryCount > 0) {
+    // eslint-disable-next-line no-use-before-define
+    return callApi(params, (retryCount - 1))
```
stcherenkov: Looks like we need `setTimeout` here

IViol: I remember, we have separate [task](https://jibrelnetwork.atlassian.net/browse/JWWEB-987) for it.

PR #1038, opened 2019-09-24
JWWEB-724 global form error

IViol:


src/pages/Send/StepOneForm.js
```diff
@@ -294,23 +310,70 @@ class StepOneForm extends PureComponent<Props, ComponentState> {
...
+    }: Props = this.props
+
+    if (nodeError || !hasInsufficientFunds) {
+      return null
+    }
+
+    return (
+      <GlobalFormError
+        text={i18n._(
+          'Send.StepOneForm.global.error.text',
```
stcherenkov: No need for `global` prefix here. You are not attaching it to any input prefix, so it is already global – related to the whole form scope.

PR #1042, opened 2019-09-27
JWWEB-797 Add wallets migration

IViol:


src/store/migrations/tests/password.test.js
```diff
@@ -0,0 +1,101 @@
...
+
+const RESULT_EMPTY = {
+  salt: '',
+  hint: '',
+  version: 1,
+  internalKey: null,
+}
+
+describe('store/migrations/password', () => {
+  test('should check if migration to v1 is needed', async () => {
```
stcherenkov: Oh, we should do a meetup on testing styles (behaviour / assert / ...)! And also decide on testing style for Wallet – I did it in another one, so I will need to refactor those a little to be consistent with yours.

src/store/migrations/db.js
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    objectStoreRequest.onerror = function objectStoreError() {
+      reject(new Error('Request failed'))
+    }
+    /* eslint-enable fp/no-mutation */
+  })
+}
+
+export function getStoreData(key: string): Promise<Object> {
```
stcherenkov: If you are interconnecting these functions with specific store format (you are suggesting that values are in JSON format), the file should not be called `db` – because these are not general functions for working with db, but specific functions for working with persist db.

There are also common functions here, like db opening, getting store version, initialising the transaction, that do not depend on actual storage structure and can be used between migration. Those are better to separate into other file for future reuse.

src/pages/WalletsMigration/WalletsMigration.js
```diff
@@ -0,0 +1,127 @@
...
+      gaSendEvent(
+        'MigrateWallets',
+        'WalletsMigrationSuccess',
+      )
+
+      window.location.reload()
+
+      return null
+    } catch (error) {
+      console.error(error)
```
stcherenkov: Let's store some important exceptions to Google Analytics, and in one of the next releases migrate it to Sentry.

GA exceptions docs: https://developers.google.com/analytics/devguides/collection/analyticsjs/exceptions

src/pages/WalletsMigration/WalletsMigration.js
```diff
@@ -0,0 +1,127 @@
...
+    : 0
+
+  return {
+    hint: passwordVersion
+      ? state.password.persist.hint
+      // $FlowFixMe
+      : (state.wallets && state.wallets.persist && state.wallets.persist.passwordOptions)
+        ? state.wallets.persist.passwordOptions.passwordHint
+        : '',
+  }
```
stcherenkov: I suggest we avoid nested ternaries whenever we can.

This one could be rewritten as:
```
if (passwordVersion) {
  return state.password.persist
}

return {
  hint: (state.wallets && state.wallets.persist && state.wallets.persist.passwordOptions)
    ? state.wallets.persist.passwordOptions.passwordHint
    : '',
}
```

src/store/migrations/v1.js
```diff
@@ -0,0 +1,232 @@
...
+    walletsStored,
+    passwordPersist,
+  )
+
+  const transaction: IDBTransaction = await initTransaction()
+
+  await deleteStoreData(TRANSACTIONS_STORE_KEY, transaction)
+  await setStoreData(walletsPersist, WALLETS_STORE_KEY, transaction)
+  await setStoreData(passwordPersist, PASSWORD_STORE_KEY, transaction)
+  await setStoreVersion(STORAGE_VERSION, transaction)
```
stcherenkov: Maybe we should wrap all of this in try..catch statement with catch sending exception to GA? Or maybe we should do it in `index.js`, so it would work the same for future migrations?

.babelrc
```diff
@@ -6,7 +6,7 @@
     "macros"
   ],
   "presets": [
-    ["@babel/env", { "modules": false }],
```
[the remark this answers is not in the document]
IViol: It didn't :D

src/AppRouter.js
```diff
@@ -68,13 +67,12 @@ class AppRouter extends Component<Props, StateProps> {
...
       // eslint-disable-next-line react/no-unused-state
       prevRouteName: null,
+      isMigrationNeeded: null,
     }
   }
 
   static getDerivedStateFromProps({ route }: Props, state: StateProps) {
-    const nextRouteName = (!route && !route.name) ?
-      constants.UNKNOWN_ROUTE :
-      route.name
```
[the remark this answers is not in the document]
IViol: I prefer `3.`. I just like to save lines in simple cases.

src/store/migrations/db.js
```diff
@@ -0,0 +1,185 @@
...
+  })
+}
+
+export function initTransaction(): Promise<IDBTransaction> {
+  return openDB().then((db: IDBDatabase) => {
+    const transaction: IDBTransaction = db.transaction(STORE_NAME, 'readwrite')
+
+    /* eslint-disable fp/no-mutation */
+    transaction.onerror = function transactionError() {
+      throw new Error('Transaction failed')
```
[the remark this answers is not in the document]
IViol: the error **stack** tells where to look for a bug :)

src/store/migrations/db.js
IViol: But `persist` container will be necessary in future as well. We need to separate data that must be persisted and data that mustn't. `persist` container is good way for this.

stcherenkov: Basically, all the data you have in storage is already persisted.

IViol: But in some reducers some data shouldn't be persisted. For that case `persist` container is ok.

IViol: Ah, ok, this container inside indexedDB is needed only when we are using redux-persist.

PR #1048, opened 2019-10-07
JWWEB-987 Update api errors management

IViol:


src/services/blockExplorer.js
```diff
@@ -265,12 +239,31 @@ function getETHTransactions(
...
+
+  const params: BlockExplorerAPIParams = {
     endblock,
     startblock,
     sort: 'desc',
     action: 'txlist',
-  }, networkId, owner)
+  }
+
+  const queryParams: string = Object
```
stcherenkov: There is a good lib called [qs](https://www.npmjs.com/package/qs) for that. It also handles edge cases and has great test coverage. Take a look.

src/services/callAPI.js
```diff
@@ -0,0 +1,48 @@
...
+  }).catch(() => handleRequestError(
+    requestInfo,
+    options,
+    params,
+    retryCount,
+  )).then((response: Response): Promise<any> => {
+    if (response.ok) {
+      return response.json()
+    }
+
```
stcherenkov: Let's also send exception to GA with future migration to Sentry. Docs on GA exceptions handling are here: https://developers.google.com/analytics/devguides/collection/analyticsjs/exceptions

PR #1058, opened 2019-10-15
JWWEB-850 Update maxLength prop for form fields

IViol:


src/components/NewPasswordForm/NewPasswordForm.js
```diff
@@ -136,6 +136,7 @@ class NewPasswordForm extends Component<Props, StateProps> {
             // eslint-disable-next-line max-len
             { defaults: 'If you forget your Security Password, some functions won’t be available. To restore access to all functions you will need to clear your data and re-import your wallets again using backup phrase.' },
           )}
+          maxLength={32}
```
stcherenkov: Too short for a hint. I suggest at least a tweet length (280 chars).

PR #1097, opened 2019-10-25
JWWEB-881 Add page with app information

IViol:


src/pages/About/About.js
```diff
@@ -0,0 +1,149 @@
...
+      })
+
+      const { i18n }: Props = this.props
+
+      this.setState({
+        version: i18n._(
+          'About.version.error',
+          null,
+          { defaults: 'App version request failed. Please, try again a bit later' },
+        ),
```
stcherenkov: Let's shorten it to just `unavailable`.

It will be enough for user to understand something went wrong, yet it takes a lot less space (and is a lot more bulletproof for translation).

src/pages/About/About.js
```diff
@@ -0,0 +1,149 @@
...
+      })
+    }
+  }
+
+  requestVersion = (): Promise<string> => {
+    if (__DEV__) {
+      return new Promise(resolve => setTimeout(() => resolve('2.2.0.691-edc1108'), 2000))
```
stcherenkov: Maybe version `develop` for `__DEV__` would be a better idea?

src/pages/About/About.js
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    return fetch(VERSION_TXT_URL).then((response: Response): Promise<string> => {
+      if (!response.ok) {
+        throw new Error('App version request failed')
+      }
+
+      return response.text()
```
stcherenkov: Let's also trim this text of spaces and line breaks.

src/pages/About/About.js
stcherenkov: (that's about text itself)

PR #1118, opened 2019-11-05
Copy lockfile during docker build

IViol:


Dockerfile
```diff
@@ -5,7 +5,7 @@ WORKDIR /app
 
 ARG NODE_ENV
 ENV NODE_ENV $NODE_ENV
-COPY package.json .
+COPY package.json package-lock.json .
```
soar:
```suggestion
COPY package*.json ./
```
Just prettier line

PR #1119, opened 2019-11-08
Add tests for merge assets function

IViol:


src/utils/digitalAssets/tests/addETHAsset.test.js
```diff
@@ -0,0 +1,26 @@
+// @flow strict
+
+import addETHAsset from '../addETHAsset'
```
stcherenkov: Better to import from `index.js` because it is the public interface you are testing for existence.

src/utils/digitalAssets/tests/addETHAsset.test.js
```diff
@@ -0,0 +1,26 @@
...
+      throw new Error('ETH asset not found')
+    }
+
+    expect(eth).toBeDefined()
+    expect(eth.symbol).toEqual('ETH')
+    expect(eth.name).toEqual('Ethereum')
+    expect(eth.blockchainParams).toBeDefined()
+    expect(eth.blockchainParams.address).toEqual(ETH_ADDRESS)
+  })
+})
```
stcherenkov: Also tests required:
- non-empty assets list
- with `ethereum` already in assets list
- with mock for `getAssetsMainnet` function returning various results (empty, non-empty, with `ethereum`)

src/utils/digitalAssets/tests/mergeDigitalAssets.test.js
```diff
@@ -0,0 +1,109 @@
...
+  test('defined', () => {
+    expect(mergeDigitalAssets).toBeDefined()
+  })
+
+  test('default items are merged correctly', async () => {
+    const result: DigitalAssets = await mergeDigitalAssets({})
+    const jnt: ?DigitalAsset = result[JNT_ADDRESS]
+
+    if (!jnt) {
+      throw new Error('JNT asset not found')
```
stcherenkov: No need to throw here.

First, you are testing `jnt` existence in next assertion.

Second, if test continues execution on assertion not passed, you could use safe assertions after, for example: https://jestjs.io/docs/en/expect#tohavepropertykeypath-value

src/utils/digitalAssets/tests/mergeDigitalAssets.test.js
```diff
@@ -0,0 +1,109 @@
...
+
+  test('existed items are merged correctly', async () => {
+    const result: DigitalAssets = await mergeDigitalAssets({
+      [AZBIT_ADDRESS]: AZBIT_ASSET,
+    })
+
+    const azbit: ?DigitalAsset = result[AZBIT_ADDRESS_CS]
+
+    if (!azbit) {
+      throw new Error('Azbit asset not found')
```
stcherenkov: Same here

src/utils/digitalAssets/tests/mergeDigitalAssets.test.js
```diff
@@ -0,0 +1,109 @@
...
+        ...AZBIT_ASSET,
+        isActive: true,
+        isCustom: false,
+      },
+    })
+
+    const azbit: ?DigitalAsset = result[AZBIT_ADDRESS_CS]
+
+    if (!azbit) {
+      throw new Error('Azbit asset not found')
```
stcherenkov: Same here

src/utils/digitalAssets/tests/mergeDigitalAssets.test.js
```diff
@@ -0,0 +1,109 @@
...
+
+  test('existed custom item will be converted to non-custom', async () => {
+    const result: DigitalAssets = await mergeDigitalAssets({
+      [JNT_ADDRESS]: AZBIT_ASSET,
+    })
+
+    const jnt: ?DigitalAsset = result[JNT_ADDRESS]
+
+    if (!jnt) {
+      throw new Error('JNT asset not found')
```
stcherenkov: Same here

src/utils/digitalAssets/tests/mergeDigitalAssets.test.js
IViol: `const eth: ?DigitalAsset = result[ETH_ADDRESS]` means that `eth` can be `null` or `undefined`.
`expect(eth).toBeDefined()` will not throw if `eth` is `null`.

